Issue #36, opened 2019-08-29
Services (Async actions / Effects)


belozer:
Example:

```ts
export type Service<T = any> = (store: Store) => (input: T) => void
```

**Example**
```ts
export const GotoPage: Service<number> = ({ getState, dispatch }) => async (page) => {
    dispatch(setPage(page))

    const api = new Api();
    const category = getState(Category);
    const res = await api.get('/products/filter', { page, category_id: category.id });

    dispatch(addProducts(res.data))
    dispatch(setProductIds(res.data.map((p: any) => p.id)))
}
```

**Usage**
```ts
const gotoPage = GotoPage(myStore);

gotoPage(10);
```

artalar:
What about
```ts
const store = createStore({
  atom: RootAtom,
  preloadedState: {},
  meta: {
    bind: action => payload => store.dispatch(action(payload)),
    network: fetch
  }
});

const requestData = declareAction(
  null,
  ({ meta: { network, bind } }) => payload =>
    network(url, payload).then(bind(responseData))
);

// domain.ts
export const declareService = reactions => store =>
  store.meta.bind(
    declareAction(null, store => payload =>
      reactions.forEach(r => r(payload, { ...store, ...store.meta }))
    )
  );
// workflow.ts
const requestData = declareService([
  (payload, { network, bind }) => network(url, payload).then(bind(responseData))
]);
```

belozer:
We need to get something like this.

```tsx
const Network = declareEnv(myDefaultFetch)
const Bind = declareEnv(myDefaultBind)

// requestData uses fetch
const store = createStore(RootAtom, {}, [
  Network(fetch) // Override defalut value in this store
]);

// requestData uses myDefaultFetch
const store2 = createStore(RootAtom, {});

// usage
const requestData = declareService([
  (payload, { getEnv }) => getEnv(Network)(url, payload).then(bind(responseData))
]);
```

belozer:
Discussion about enviroment of store #38

PR #37, opened 2019-08-29
core/types: add Service type

belozer:
#36

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -395,6 +395,9 @@ export type Store = {
   getState: typeof storeGetState
 }
 
+// Use for async operations and compose dispatch calls in one function
+export type Service<T = any> = (store: Store) => (input: T) => void
+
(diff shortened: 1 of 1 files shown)


artalar:
What about
```ts
const store = createStore({
  atom: RootAtom,
  preloadedState: {},
  meta: {
    bind: action => payload => store.dispatch(action(payload)),
    network: fetch
  }
});

const requestData = declareAction(
  null,
  ({ meta: { network, bind } }) => payload =>
    network(url, payload).then(bind(responseData))
);

// domain.ts
export const declareService = reactions => store =>
  store.meta.bind(
    declareAction(null, store => payload =>
      reactions.forEach(r => r(payload, { ...store, ...store.meta }))
    )
  );
// workflow.ts
const requestData = declareService([
  (payload, { network, bind }) => network(url, payload).then(bind(responseData))
]);
```

PR #47, opened 2019-09-02
types improvements

Wroud:
change actionTypes type from `Map<true>` to `Array<string>`
change some types, add `Dictionary<T>` and replace `Record<T>` with them
make `createStore` first argument optional
replace `throwIf` with `throwError`

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -3,7 +3,10 @@
   "version": "0.1.0",
   "description": "React bindings for @reatom/core",
   "homepage": "https://github.com/artalar/reatom/tree/master/packages/react",
-  "repository": "https://github.com/artalar/reatom",
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/artalar/reatom.git"
+  },
   "license": "MIT",
   "source": "src/index.ts",
   "main": "build/index.js",


src/index.ts
```diff
@@ -2,8 +2,8 @@ console.warn('REAtom still work in progress, do not use it in production')
 
 type NodeId = string
 
-type ActionType = string
-type ActionTypesDictionary = Record<ActionType, true>
+type Dictionary<T> = { [key: string]: T }
+type ActionTypesDictionary = Record<NodeId, true>
```
artalar: ` type ActionTypesDictionary = Record<ActionType, true>` was right

src/index.ts
```diff
@@ -2,8 +2,8 @@ console.warn('REAtom still work in progress, do not use it in production')
 
 type NodeId = string
 
-type ActionType = string
-type ActionTypesDictionary = Record<ActionType, true>
+type Dictionary<T> = { [key: string]: T }
```
artalar: ` type Dictionary<T> = { [key: string]: T | undefined }`

src/index.ts
artalar: But you may replace Record by Dictionary

Wroud: mean like `type ActionTypesDictionary = Dictionary<bool>`?

src/index.ts
```diff
@@ -46,7 +46,7 @@ let id = 0
 function nameToId(name: unknown) {
   return Array.isArray(name)
     ? safetyStr(name[0], 'name')
-    : safetyStr(name, 'name') + ' #' + ++id
+    : `${safetyStr(name, 'name')} #${++id}`
```
belozer: don't. It's a very slow operation.

src/index.ts
```diff
@@ -155,7 +155,7 @@ export function declareAtom<State>(
 
   throwIf(
     initialState === undefined,
-    'Atom "' + atomId + '". Initial state can\'t be undefined',
+    `Atom "${atomId}". Initial state can't be undefined`,
```
belozer: don't. It's a very slow operation.

src/index.ts
```diff
@@ -220,13 +220,7 @@ export function declareAtom<State>(
...
         throwIf(
           atomStateNew === undefined,
-          'Invalid state. Reducer №' +
-          position +
-          ' in ' +
-          '"' +
-          atomId +
-          '"' +
-          ' atom returns undefined',
+          `Invalid state. Reducer № ${position} in "${atomId}" atom returns undefined`,
```
belozer: don't. It's a very slow operation.

src/index.ts
```diff
@@ -336,7 +330,7 @@ export function combine(name: any, shape: any) {
   if (arguments.length === 1) {
     // @ts-ignore
     shape = name
-    name = '{' + (keys = Object.keys(shape)).join() + '}'
+    name = `{${(keys = Object.keys(shape)).join()}}`
```
belozer: don't. It's a very slow operation.

src/index.ts
Wroud: its exception....

belozer: It does not matter :) it is Better to do work with strings in the same style.

Wroud: but code will run anyway so better implement string formatting in `throwIf`  if we want better perfomace

belozer: But for exceptions... Maybe we should talk about it.

belozer: About better perfomance. See https://gist.github.com/c564436f39a8a1557b5969f405ccb492

belozer: rather it is http://jsbench.github.io/#c564436f39a8a1557b5969f405ccb492

src/index.ts
```diff
@@ -524,21 +516,30 @@ function callFromList(list: Function[], arg: any, i = -1) {
   while (++i < list.length) list[i](arg)
 }
 
+function formatString(message: string, ...args: any[]) {
```
belozer: Loss in performance and extra weight to the package.

package.json
```diff
@@ -18,11 +18,14 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-typescript": "^7.3.3",
     "@types/jest": "^24.0.18",
+    "@types/react": "^16.8.0",
```
belozer: it's not necessary here.

belozer: react and redux too

belozer: your need run `npm i` from `packages/packageName/`. Adding them here is not the solution.

lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "hoist": true,
+  "hoist": false,
```
belozer: you're working with true, right?

Wroud: yep

belozer: so it's not necessary here

Wroud: its kind are bug, this change was mark outdated

belozer: ohhh.... Sorry

PR #70, opened 2019-09-07
Types cleanup

Wroud:


packages/core/src/index.ts
```diff
@@ -1,37 +1,38 @@
 console.warn('REAtom still work in progress, do not use it in production')
 
-type NodeId = string
-
-type ActionType = string
```
artalar: Please, return ActionType and NodeId. It will helpful with nominal types (https://twitter.com/WrocTypeScript/status/1168498435110121472?s=20)

Wroud: [Here](https://github.com/artalar/reatom/blob/1f740b64932658c257da41a64eb5e28160629bb2/packages/core/src/index.ts#L92) conflict with type cast from `NodeId` to `ActionType`

packages/core/src/index.ts
```diff
@@ -1,37 +1,38 @@
...
+) => void
 
 const NODE = Symbol('@@REAtom/NODE')
 const assign = Object.assign
 
-function noop() {}
+function noop() { }
 
-export type ActionCreator<Payload = undefined, Type extends string = string> = {
+export type ActionCreator<TPayload = undefined, TType extends string = string> = {
```
belozer: Why not `Payload` and `Type`?

Wroud: When generic types uses without prefix it may result with type overrides 
```suggestion
export declare function declareAtom<State>(
export declare function declareAtom<TState>(
```
[its example](https://github.com/artalar/reatom/pull/70/files#diff-55646d75364ce42cdb84abec1a1531ccR120) when we have type `State` declared in file and generic type `State` declared in function

Wroud: it's more fun when we get something like this:
```javascript
export declare function declareAtom<State extends State>(
```

belozer:
```
export declare function declareAtom<S extends State>(
```

belozer: @Wroud see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L114

belozer: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L119

Wroud: it's good way too

PR #77, opened 2019-09-10
core/logic/add: new kernel

artalar:


packages/core/src/shared.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function throwError(error: string) {
+  // TODO: add link to docs with full description
+  throw new Error('[reatom] ' + error)
+}
+export function safetyStr(str: string, name: string): string {
+  if (typeof str !== 'string' || str.length === 0) throwError(`Invalid ${name}`)
+  return str
+}
+export function safetyFunc(func: unknown, name: string) {
```
belozer: @artalar why not `export function safetyFunc(func: Function, name: string) {` ?

PR #81, opened 2019-09-11
refactor(core): many type fixes and improvements

Wroud:
add `Reducer` type
add `Reduce` type
add `DependencyMatcher`
add `T` prefix for generics
add `Dictionary` type
rename type `Fn` to `AtomUpdater`
rename `Ctx.changedIds` to `changedTreeIds`
replace `ActionType` with `Leaf`
make `Ctx` type static declared and provide `TPayload` generic
fix type guards `getTree`, `getIsAtom`, `getIsAction`


Wroud:
close until better times...

Issue #52, opened 2019-09-03
Generate unit prefix by LoC


artalar:
We need the option to replace (at runtime) the unique name generator (`nameToId`) for adding postfix based on a line of code (it is possible to define in env at V8 by error instance stack).

[Tip](https://stackoverflow.com/questions/13227489/how-can-one-get-the-file-path-of-the-caller-function-in-node-js)

Example usage

```js
// shared.ts
let nameToId = () => { /*...*/ }
export function replaceNameToId = _nameToId => nameToId = _nameToId

// user-space.ts
import { replaceNameToId } from '@reatom/core'
if (process.env.DEV) replaceNameToId(
  // https://stackoverflow.com/questions/13227489/how-can-one-get-the-file-path-of-the-caller-function-in-node-js
)
```

artalar:
done

PR #101, opened 2019-09-17
feat: implement setNameToId #52

Wroud:


packages/core/src/__tests__/shared.ts
```diff
@@ -0,0 +1,16 @@
...
+  describe('shared', () => {
+    test('setNameToId', () => {
+      // @ts-ignore
+      setNameToId(name => name + 10)
+      const at = declareAtom('pep', {}, () => {})
+      const act = declareAction('peps')
+
+      expect(at[TREE].id).toBe('pep10')
+      expect(act[TREE].id).toBe('peps10')
+      expect(nameToId('a')).toBe('a10')
```
artalar: it should be the first test

packages/core/src/__tests__/shared.ts
```diff
@@ -0,0 +1,16 @@
+import { declareAction, declareAtom, setNameToId } from '../index'
+import { nameToId, TREE } from '../shared'
+describe('@reatom/core', () => {
```
artalar: add `\n` please :)

Issue #120, opened 2019-09-21
We need massive refactoring and code cleanup, typings rework


Wroud:
Current typings implementation is broken inside Reatom

PR #121, opened 2019-09-21
refactor(core): typings and code cleanup

Wroud:
Part of #120

### DIFF
--- a/packages/core/src/__tests__/index.ts
+++ b/packages/core/src/__tests__/index.ts
@@ -360,13 +360,13 @@ describe('@reatom/core', () => {
       expect(store.getState(count)).toBe(4)
       expect(store.getState().countDoubled).toBe(8)
     })
-    test('createStore with nullable atom', () => {
+    test('createStore with undefined atom', () => {
       const increment = declareAction()
       const countStatic = declareAtom(['countStatic'], 0, handle =>
         handle(increment, state => state + 1),
       )
 
-      const store = createStore(null, { countStatic: 10 })
+      const store = createStore(undefined, { countStatic: 10 })
       store.dispatch(increment())
 
       expect(store.getState(countStatic)).toBe(10)
--- a/packages/core/src/createStore.ts
+++ b/packages/core/src/createStore.ts
@@ -1,18 +1,7 @@
-import { Leaf, Tree, State, TreeId, Ctx, createCtx } from './kernel'
-import {
-  TREE,
-  noop,
-  nameToId,
-  Unit,
-  throwError,
-  getTree,
-  safetyFunc,
-  getIsAction,
-  assign,
-  getIsAtom,
-} from './shared'
-import { Action, declareAction, ActionType } from './declareAction'
-import { Atom, declareAtom, initAction, map, getState } from './declareAtom'
+import { Tree, State, TreeId, createCtx } from './kernel'
+import { throwError, getTree, safetyFunc, assign, getIsAtom } from './shared'
+import { Action } from './declareAction'
+import { Atom, declareAtom, initAction, getState } from './declareAtom'
 
 type DispatchFunction = (action: Action<any>) => any
 type SubscribeFunction = {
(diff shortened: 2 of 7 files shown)


packages/core/src/createStore.ts
```diff
@@ -1,26 +1,15 @@
...
+import { Atom, declareAtom, initAction, getState } from './declareAtom'
 
 type DispatchFunction = (action: Action<any>) => any
 type SubscribeFunction = {
   <T>(target: Atom<T>, listener: (state: T) => any): () => void
   (listener: DispatchFunction): () => void
 }
 type GetStateFunction = {
-  <T>(target: Atom<T>): T
+  <T>(target: Atom<T>): T | undefined
```
belozer: Why undefined?

belozer: GetState always returns state of atom in store or initial state

packages/core/src/createStore.ts
```diff
@@ -68,19 +57,20 @@ export function createStore(
     }
   }
 
-  function _getState(target?: Atom<any>) {
+  function _getState(): State
+  function _getState<T>(target?: Atom<T>): State | T | undefined {
```
belozer: _getState not retruns `undefined`

packages/core/src/createStore.ts
Wroud: it's related with `getState` function that can return undefined

packages/core/src/createStore.ts
Wroud: if so we need change type of `getState` from `declareAtom.ts`

packages/core/src/declareAtom.ts
```diff
@@ -74,8 +59,8 @@ export function declareAtom<State>(
     throwError(`Atom "${_id}". Initial state can't be undefined`)
 
   function reduce<T>(
-    dep: Unit<T>,
-    reducer: (state: State, payload: T) => State | undefined,
+    dep: Unit,
+    reducer: (state: TState, payload: T) => TState | undefined,
```
belozer: @artalar this is bug? `undefined` not alowed for state

packages/core/src/declareAtom.ts
```diff
@@ -136,13 +121,10 @@ export function declareAtom<State>(
     }
   }
 
-  reduce(_initAction, (state = initialState) => state)
+  reduce(_initAction, (state = initialState as TState) => state)
```
belozer: Why not? `(state: TState = initialState) => state`

Wroud: here we need cast because initialState argument is union

packages/core/src/declareAtom.ts
```diff
@@ -160,66 +140,64 @@ export function declareAtom<State>(
   return atom
 }
 
-export function getState<T>(state: State, atom: Atom<T>): T | undefined {
+export function getState<T>(state: State, atom: Atom<T>): T {
```
artalar: getState and store.getState is different functions. The getState must return optional type

packages/core/src/declareAtom.ts
```diff
@@ -160,66 +140,64 @@ export function declareAtom<State>(
...
   return state[atom[TREE].id]
 }
 
-// @ts-ignore
-export declare function map<T, _T = unknown>(
+export function map<T, _T = unknown>(
```
artalar: Type inference for the map is broking
proof

packages/core/src/__tests__/index.ts
```diff
@@ -360,13 +360,13 @@ describe('@reatom/core', () => {
...
     })
-    test('createStore with nullable atom', () => {
+    test('createStore with undefined atom', () => {
       const increment = declareAction()
       const countStatic = declareAtom(['countStatic'], 0, handle =>
         handle(increment, state => state + 1),
       )
 
-      const store = createStore(null, { countStatic: 10 })
+      const store = createStore(undefined, { countStatic: 10 })
```
belozer: :confused: why not `null`? the code looks strange in this case

Wroud: Function with null, cant have default value

belozer: It's not like the usual code in all libraries.

Wroud: You should not use null as undefined in javascript, null is value, but we used it as no value provided, we can't use default argument value for that

PR #138, opened 2019-09-27
Fix/124

artalar:


packages/react/README.md
```diff
@@ -13,40 +13,54 @@ Package for bindings [Reatom](https://github.com/artalar/reatom) store with Reac
...
+
 ```ts
-useAtom(atom: Atom, silent: boolean = true): AtomState
+const atomValue = useAtom(atom)
+```
+
+#### Depended value by selector
+
+```ts
+const atomValue = useAtom(atom, atomState => atomState[props.id], [props.id])
```
Wroud: Can props been changed before callback was called?

packages/react/src/index.ts
```diff
@@ -19,64 +20,91 @@ function useForceUpdate() {
...
+export function useAtom<T>(atom: Atom<T>): T
+export function useAtom<TI, TO = TI>(
+  atom: Atom<TI>,
+  mapper: (atomValue: TI) => TO,
+  deps: any[],
+): TO
+export function useAtom<TI, TO = TI>(
+  atom: Atom<TI>,
+  mapper: (atomValue: TI) => TO = (atomValue: TI) =>
+    (atomValue as unknown) as TO,
```
Wroud: mapper and deps must be optional here

Wroud: ... its already optional, sorry 🙏

packages/react/src/__tests__/index.tsx
```diff
@@ -1,23 +1,193 @@
...
+  )
 }
 
 describe('@reatom/react', () => {
-  test('should throw if provider is not set', () => {
-    expect(() => render(<Counter />)).toThrowError(
-      '[reatom] The provider is not defined',
-    )
+  describe('useAtom', () => {
+    test('throw Error if provider is not set', () => {
```
belozer: What's about useAction?

packages/react/src/index.ts
```diff
@@ -19,64 +20,91 @@ function useForceUpdate() {
...
+  deps: any[] = [],
```
belozer: `deps?: any[] = [],`

packages/react/src/index.ts
```diff
@@ -19,64 +20,91 @@ function useForceUpdate() {
...
-): T {
-  const atomRef = useRef<Atom<T>>()
-  const stateRef = useRef<T>()
-  const unsubscribeRef = useRef<Function>()
-  const isMountRef = useRef(false)
+export function useAtom<T>(atom: Atom<T>): T
+export function useAtom<TI, TO = TI>(
+  atom: Atom<TI>,
+  mapper: (atomValue: TI) => TO,
+  deps: any[],
```
belozer: `deps?: any[] = [],`

packages/react/src/index.ts
artalar: An argument with default value is optional (by typings) as is

packages/react/src/index.ts
artalar: An argument with default value is optional (by typings) as is

PR #162, opened 2019-10-14
refactor

artalar:


packages/core/src/__tests__/index.ts
```diff
@@ -54,80 +54,95 @@ describe('@reatom/core', () => {
...
+        )
       })
-    })
-    test('throw error if declareAtom called with undefined initial state', () => {
-      const run = () => declareAtom(['test'], undefined, reduce => [])
+      test('throw error if atom produced undefined value', () => {
+        const action = declareAction()
+
+        expect(() =>
+          declareAtom(['myAtom'], {}, r => r(action, () => undefined as any))(
```
belozer: The current file uses `reduce`, `r` and `on`. The desired uniformity so as not to drive you crazy readers.

belozer: I suggest to fix `on`

packages/core/src/createStore.ts
```diff
@@ -28,25 +32,31 @@ export function createStore(atom: Atom<any>, initState?: State): Store
...
+        assign(state, initState)
+      else if (initState !== undefined) throwError('Invalid initial state')
+
+      storeTree.union(getTree(atom as Atom<any>))
+      const ctx = createCtx(state, initAction)
+      storeTree.forEach(initAction.type, ctx)
+      assign(state, ctx.stateNew)
+      initialAtoms = new Set(Object.keys(ctx.stateNew))
+    }
+  }
```
Wroud: I think there can be more simple than tons of 'if'

packages/core/src/declareAtom.ts
```diff
@@ -121,8 +121,8 @@ export function declareAtom<TState>(
     depTree.fnsMap.forEach((_, key) => _tree.addFn(update, key))
   }
 
-  handler(_initAction, (state = initialState as TState) => state)
-  dependencyMatcher(handler)
+  on(_initAction, (state = initialState as TState) => state)
+  dependencyMatcher(on)
```
Wroud: Maybe keep handler? it is more descriptive

artalar: `on` is friendly for new contributors

belozer:
fix(core): naming in tests

fix - needs use only for fix bugs. In this case needs use one of `style`, `refactor`, or `chore`

packages/core/src/createStore.ts
artalar: let propose a changes in a new PR

PR #171, opened 2019-10-15
declareAction: fix validating in the second generic

### LINKED ISSUE #169: Action type generic is not validating arguments
Second generic must be valid only if _type_ set: `declareAction<any, 'action name'>([''action name''])`

example

ilyaryabchinski:
may close #169

### DIFF
--- a/packages/core/src/declareAction.ts
+++ b/packages/core/src/declareAction.ts
@@ -48,7 +48,7 @@ export function declareAction<
   Payload = undefined,
   Type extends ActionType = string
 >(
-  name: string | [Type] | Reaction<Payload> = 'action',
+  name: string | [Type] | Reaction<Payload>,
   ...reactions: Reaction<Payload>[]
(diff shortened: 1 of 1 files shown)


belozer:
@ilyaryabchinski problem with build

```
(rpt2 plugin) Error: /home/runner/work/reatom/reatom/packages/core/src/declareAction.ts(27,17): semantic error TS2394 This overload signature is not compatible with its implementation signature.
Error: /home/runner/work/reatom/reatom/packages/core/src/declareAction.ts(27,17): semantic error TS2394 This overload signature is not compatible with its implementation signature.
    at error (/home/runner/work/reatom/reatom/node_modules/rollup/dist/rollup.js:3460:30)
    at Object.error (/home/runner/work/reatom/reatom/node_modules/rollup/dist/rollup.js:21251:17)
```

artalar:
Fixed in `@reatom/core@1.0.0-rc18`

PR #227, opened 2019-10-28
refactor(react): improve useForceUpdate hook

jeetiss:
`useReduser` is better solution for force update hook: 
- react guarantees that dispatch function identity is stable and won’t change on re-renders
- with reducer `v => !v` state will change every dispatch and trigger re-render

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -2,10 +2,10 @@ import {
   createContext,
   useEffect,
   useRef,
-  useState,
+  useReducer,
   useContext,
   useCallback,
-  MutableRefObject,
+  Reducer,
   useMemo,
 } from 'react'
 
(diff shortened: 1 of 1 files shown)


packages/react/src/index.ts
```diff
@@ -22,8 +22,9 @@ function noop() {}
 export const context = createContext<Store | null>(null)
 
 function useForceUpdate() {
-  const update = useState(0)[1]
-  return useRef(() => update(v => v + 1)).current
+  const [_, update] = useReducer<Reducer<boolean, null>>(s => !s, false)
+  // update don't have action and don't changes between rerenders
+  return update as () => void
```
artalar:
```suggestion
  return useReducer<Reducer<boolean, null>>(s => !s, false)[1] as () => void
```

jeetiss:
Woow, github can't commit multiline suggestions
I fixed this

artalar:
@all-contributors please add @jeetiss  for code

PR #239, opened 2019-11-03
feat(react): export Provider component instead of all context

### LINKED ISSUE #238: @reatom/react package should export Provider only instead of context

Kurzdor:
fixes #238


packages/react/src/index.ts
```diff
@@ -19,7 +19,11 @@ import {
 
 function noop() {}
 
-export const context = createContext<Store | null>(null)
+const context = createContext<Store | null>(null)
```
belozer: Breaking change. This code is already used in projects.

```
export const context = createContext<Store | null>(null)
```

belozer:
Needs rename commit from `refactor(react)` to `feat(react)` bacause this adding new feature

Kurzdor:
now non-breakable change

PR #258, opened 2019-12-15
fix(debug): make config optional

arswarog:


artalar:
@allcontributors please add @arswarog for code

PR #282, opened 2020-02-01
feat: better context and store provider (breaking change)

Kurzdor:
It's a breaking change.

1. StoreProvider instead of Provider
2. Updated docs and tests to new StoreProvider change
3. Refactored context to separate file
4. Also, bundles jsx


artalar:
@Kurzdor may you describing your motivation?
[Redux using](https://github.com/reduxjs/react-redux/blob/master/src/index.js#L17) just `Provider`

artalar:
Kurzdor, if you want to finish this changes, please, create an issue and let discuss it.

PR #286, opened 2020-02-05
feat(future): init and implementation

artalar:


packages/future/tests/index.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const cleanup = jest.fn((id: number) => clearInterval(id))
+    const cb = jest.fn((v: number) => v === 2 && resolve())
+    const unsubscribe = f.subscribe(cb)
+
+    let resolve!: Function
+    await new Promise(r => resolve = r)
+    unsubscribe()
+
+    expect(cb).toBeCalledTimes(2)
+    expect(cleanup).toBeCalledTimes(1)
```
belozer: This is test for case with one subscrpition. But needs test multiple subscriptions.

PR #296, opened 2020-02-25
chore: eslint configuration setup

stenin-nikita:


artalar:
Awesome! Thx!
@allcontributors please add @stenin-nikita for infra

PR #298, opened 2020-02-29
fix(core): fix reducers result equality

### LINKED ISSUE #291: Fix reducers result equality
- [ ] New state must be take only if it changed ([here](https://github.com/artalar/reatom/blob/99ae221d412c7e81405950f6468de5a7fb9ff9ab/packages/core/src/declareAtom.ts#L124))
- [ ] Use `Object.is` instead of `===`
- [ ] If `hasAtomNewState` and `atomStateNew === atomStateSnapshot` `changedIds` must not contains atom id.

stenin-nikita:
fix #291

### DIFF
--- a/packages/core/src/declareAtom.ts
+++ b/packages/core/src/declareAtom.ts
@@ -10,6 +10,7 @@ import {
   getIsAction,
   assign,
   getName,
+  equals,
 } from './shared'
 import { Action, declareAction, PayloadActionCreator } from './declareAction'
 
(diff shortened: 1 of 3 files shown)


packages/core/tests/index.test.ts
```diff
@@ -685,4 +685,49 @@ describe('@reatom/core', () => {
...
+
+    const store = createStore(counterAtom)
+    expect(counterReducerMock).toBeCalledTimes(1)
+
+    store.dispatch(action())
+    expect(counterReducerMock).toBeCalledTimes(2)
+
+    store.dispatch(action())
+    expect(counterReducerMock).toBeCalledTimes(2)
+  })
```
artalar: If `hasAtomNewState` and `atomStateNew === atomStateSnapshot` `changedIds` must not contains atom id.
Try this case:
```js
declareAtom(on => [
  on(action, s => s + 1),
  on(action, s => s - 1)
])
```

packages/core/src/declareAtom.ts
```diff
@@ -120,8 +121,18 @@ export function declareAtom<TState>(
             `Invalid state. Reducer number ${position} in "${_name}" atom returns undefined`,
           )
 
-        if (atomStateNew !== atomState && !hasAtomNewState) changedIds.push(_id)
-        stateNew[_id as string] = atomStateNew
+        if (hasAtomNewState && equals(atomStateSnapshot, atomStateNew)) {
+          changedIds.splice(changedIds.indexOf(_id), 1)
+          delete stateNew[_id as string]
```
artalar: May be it is unnecessary? Subscribers calls by changedIds and assign equal newState to state is do nothing?

stenin-nikita: May be, but then the dependency will be marked changed for the subscribers.
https://github.com/artalar/reatom/blob/dc1b26b214f8bf34f8fb9b5fdb53196c5702889e/packages/core/src/declareAtom.ts#L111-L112

PR #302, opened 2020-03-01
fix(core): improve type inference with TypeScript 3.5

### LINKED ISSUE #301: Incorrect type inference with TypeScript 3.5
How to reproduce:
```typescript
import { declareAtom, declareAction } from '@reatom/core'

const action = declareAction()

export const atom = declareAtom({ foo: 'bar' }, on => [
  on(action, () => ({ foo: 'bar' })),
])

export const subAtom = declareAtom(null, on => [
  // atom => Atom<{ foo: string }>
  on(atom, (_, payload) => {
    // payload => Record<TreeId, unknown> <- incorrect type
    return payload
  }),
])
```

stenin-nikita:
fix #301

### DIFF
--- a/packages/core/src/declareAtom.ts
+++ b/packages/core/src/declareAtom.ts
@@ -23,12 +23,20 @@ export const initAction = initActionCreator()
 type AtomName = TreeId | [string]
 type AtomsMap = { [key: string]: Atom<any> }
 type Reducer<TState, TValue> = (state: TState, value: TValue) => TState
-type DependencyMatcher<TState> = (
-  on: <T>(
+
+/**
+ * This interface needed for correct type inference with TypeScript 3.5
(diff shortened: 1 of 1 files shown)


packages/core/src/declareAtom.ts
```diff
@@ -22,12 +22,15 @@ export const initAction = initActionCreator()
 type AtomName = TreeId | [string]
 type AtomsMap = { [key: string]: Atom<any> }
 type Reducer<TState, TValue> = (state: TState, value: TValue) => TState
-type DependencyMatcher<TState> = (
-  on: <T>(
+interface DependencyMatcherOn<TState> {
```
artalar: Add a comment to the issue, pls (why we use an interface here)

stenin-nikita: Ok, fixed

PR #318, opened 2020-03-14
Remove irrelevant keywords from package.json

risenforces:
`effector` and `mobx` keywords are irrelevant in the context of `reatom`, so they should be removed from there.


artalar:
Please, follow the https://github.com/artalar/reatom/blob/master/CONTRIBUTING.md#sending-a-pull-request (targeting `next` branch)

PR #328, opened 2020-04-18
Issue 208 automatically generate markdown documentation from packages

ivan-nemtinov:
See docs/contributing.md ## API docs for instructions.

The main idea is to have single source of truth about documentation. 

Every generated package doc is started from README.md and completed with the API docs, automatically generated from sources.

The API docs are taken from https://github.com/artalar/reatom/pull/236 by @belozer

The docs are incomplete and wrong somewhere, but first of all I did proof of concept and the docs can be fixed in additional commits.

Also changed some existing files like CONTRIBUTING.md from root to point to docs/contributing.md (should change the name to uppercase later).

If you see an error
TypeError: Cannot read property 'slice' of undefined
at LernaPackagesPlugin.onBeginResolve
delete package_lock.json and node_modules and run npm i. Also see https://github.com/marcj/typedoc-plugin-lerna-packages/issues/21


ivan-nemtinov:
The TypeDoc is pretty hard to customize (at least, a fork is required), so I had to create a script in scripts/docs-fix to fix some parts with regexes. I've no possibility to make it better now.

ivan-nemtinov:
Also I had to mark some code with @internal to exclude these from the generated docs.
I think, some names are not looking good (for example, Ctx). Is it possible to change such names to be more meaningful?

PR #352, opened 2020-10-01
fix(react): add batched wrapper for useAction

### LINKED ISSUE #351: Cannot update state inside setInterval
Hello
I have to change the atom value inside setInterval function. 

My first version of the code:
https://gist.github.com/defint/3faffb974ab76b1f7aa8adf0ff2ebadb 
It doesn't work.

My second version of the code with auto-clicking on the button:
https://gist.github.com/defint/fcf383dda69a934bfa411c2b89b72bc4
It works fine.

Can you please describe how to change the state without auto-clicking?

rrrshtt:
fix #351

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -14,12 +14,6 @@ import ReactDOM from 'react-dom'
 
 function noop() {}
 
-function batchedFunction(func: Function) {
-  return (...funcArgs: any[]) => {
-    ReactDOM.unstable_batchedUpdates(() => func(...funcArgs))
-  }
-}
-
 export const context = createContext<Store | null>(null)
 
 export const { Provider } = context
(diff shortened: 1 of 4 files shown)


packages/react/src/index.ts
```diff
@@ -10,9 +10,16 @@ import {
 } from 'react'
 
 import { Store, Atom, Action } from '@reatom/core'
+import ReactDOM from 'react-dom'
 
 function noop() {}
 
+function batchedFunction(func: Function) {
```
artalar: The utility function is unnecessary.

packages/react/src/index.ts
```diff
@@ -141,16 +148,17 @@ export function createActionHook(ctx: Context<Store | null> = context) {
...
     if (!store) throw new Error('[reatom] The provider is not defined')
     if (typeof cb !== 'function') {
       throw new TypeError('[reatom] `useAction` argument must be a function')
     }
 
-    return useCallback((...args) => {
-      const action = cb(...args)
-      if (action) store.dispatch(action)
-    }, deps.concat(store))
+    return batchedFunction(
```
artalar: Here u neutralizing the useCallback effect and return a new function every time. It will be better to wrap the dispatch to unstable_batchedUpdates

artalar (changes requested):

artalar:
@rrrshtt thanks a lot for u work on this. I leave some comments, fix it please and add some tests.

artalar (approved):

Issue #365, opened 2021-07-04
Not work in react-native


BataevDaniil:
Reatom have in peerDependencies react-dom. I want to have not binding to react-dom

artalar:
Khm, good question...
React@18 incudes needed features and don't require react-dom any more. So I may publish a next majore version of the reatom/react package with react@18 in requirements. But I can't do it becouse the the next version already employed by _the next majore version_ with new features (reatom/core@2).

I need more time to investigate this problem.

BataevDaniil:
Я suppose the issue can be solved by adding a separate reactive-native file and linking its in package.json as followed: https://github.com/styled-components/styled-components/blob/main/packages/styled-components/package.json#L7. This file will conclude temporary fix for react-native.

Currently, I have to mock react-dom with local library containing only one function which just invokes everything what was passed to it.
````js
exports.default = {
  unstable_batchedUpdates: (f) => {
    f();
  },
};
````
I'm worried to use that way because it definitely can lead to some bugs.
What do you think of that temporary solution?
I can create a PR if you don't mind.

Thank you for your activity)

artalar:
Probably yes, thx for example link. I try to fix it tomorrow.

artalar:
@BataevDaniil sry, I was inattentive at the last time.

> What do you think of that temporary solution?

It is ok.

> I can create a PR if you don't mind.

I will really pleasure for it.

> Thank you for your activity)

Thanks for your report and solution suggestion.

Also, check this out https://github.com/alloc/react-batched-updates

BataevDaniil:
Solution work for reatom v1 and [here](https://github.com/artalar/reatom/pull/377) solution for reatom v2

PR #366, opened 2021-07-22
feat(react): add support react-native 

BataevDaniil:
https://github.com/artalar/reatom/issues/365


artalar:
This is not a best solution, because it considering only two options for React backend (DOM and _native_), but it could me more.
Anyway, my thought is about to accept this and expect no new same issues with react@18, which is close enought.

BataevDaniil:
>  it considering only two options for React backend (DOM and native), but it could me more.

Yes, for each platform need own solution (adapter for each platform). Now adapter is just if.

PR #377, opened 2021-09-16
Fix import memo and add support batchUpdated

BataevDaniil:


artalar:
@allcontributors please add @BataevDaniil for code

PR #393, opened 2021-11-02
defaultStore option added

### LINKED ISSUE #392: Allow to change defaultStore in createAtom
We need an option `defaultStore` in [AtomOptions](https://github.com/artalar/reatom/blob/v2/packages/core/src/createAtom.ts#L46). With this a user may create own createAtom implementation with own store, which useful for its customization:

```ts
// utils/store.ts
export const store = createStore({
  onPatch: console.log
})

// utils/createAtom.ts
import {createAtom as createAtomOriginal} from '@reatom/core'
export const createAtom: typeof createAtomOriginal = (deps, reducer, options) =>
  createAtomOriginal(deps, reducer, {...options, defaultStore: store})
```

Akiyamka:
closes #392

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -451,7 +451,13 @@ const formAtom = createAtom(
 
 > `Outdated track call` is throwed when you try to use reactive handlers outside synchronous reducer call. For example, `schedule` is called only after all atoms recalculation.
 
-> TODO: `options`
+The third argument `options` allows you to override the default atom settings:
+id - setup id for this atom
+decorators - array of (atom decorators)[#atom-decorators] 
+defaultStore - redefine store for this atom
+
+### Atom decorators 
+> TODO: Add example
 
 ## Internal architecture
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -454,7 +454,7 @@ const formAtom = createAtom(
 The third argument `options` allows you to override the default atom settings:
 id - setup id for this atom
 decorators - array of (atom decorators)[#atom-decorators] 
-defaultStore - redefine store for this atom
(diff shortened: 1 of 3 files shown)


packages/core/README.md
```diff
@@ -451,7 +451,13 @@ const formAtom = createAtom(
 
 > `Outdated track call` is throwed when you try to use reactive handlers outside synchronous reducer call. For example, `schedule` is called only after all atoms recalculation.
 
-> TODO: `options`
+The third argument `options` allows you to override the default atom settings:
+id - setup id for this atom
+decorators - array of (atom decorators)[#atom-decorators] 
+defaultStore - redefine store for this atom
```
artalar:
```suggestion
defaultStore - redefine store bindings for `dispatch`, `getState` and `subscribe` of this atom
```

packages/core/src/createAtom.ts
```diff
@@ -9,16 +9,16 @@ import {
   Cause,
   createReatomError,
   createTemplateCache,
-  defaultStore,
+  defaultStore as originalStore,
```
artalar:
```suggestion
  defaultStore as originalDefaultStore,
```

artalar: Also, if you can, rename please `callSafetyDefault` to `originalCallSafety` and `createTemplateCacheDefault` to `originalCreateTemplateCache` [here](https://github.com/artalar/reatom/blob/v2/packages/core/src/createStore.ts)

artalar (changes requested):

packages/core/src/createAtom.ts
Akiyamka: done

artalar:
@allcontributors please add @Akiyamka for code

PR #395, opened 2021-11-13
fix default store

BataevDaniil:


artalar:
This solution have some pros and cons and I can't approve it now, we need to think about it a little more

PR #400, opened 2022-02-23
Update createPrimitiveAtom.ts

Akiyamka:
Store options ignored in createPrimitiveAtom

### DIFF
--- a/packages/core/primitives/createPrimitiveAtom.ts
+++ b/packages/core/primitives/createPrimitiveAtom.ts
@@ -99,7 +99,7 @@ export function createPrimitiveAtom<State>(
 
     noop,
 
-    { decorators, store, id },
+    { decorators, ...restOptions },
   )
 
   const actionCreatorsTypes = keys.map((key) => atom[key].type)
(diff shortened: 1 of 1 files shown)


packages/core/primitives/createPrimitiveAtom.ts
```diff
@@ -60,7 +60,7 @@ export function createPrimitiveAtom<State>(
       payload(state),
   }
 
-  let { decorators = [], id } = isString(options)
+  let { decorators = [], store, id } = isString(options)
```
artalar:
```suggestion
  let { decorators = [], ...restOptions } = isString(options)
```

packages/core/primitives/createPrimitiveAtom.ts
```diff
@@ -99,7 +99,7 @@ export function createPrimitiveAtom<State>(
 
     noop,
 
-    { decorators, id },
+    { decorators, store, id },
```
artalar:
```suggestion
    { decorators, ...restOptions },
```

PR #418, opened 2022-05-28
Add whatsup to benchmarks

iminside:


artalar:
Added here https://github.com/artalar/reactive-computed-bench/commit/9537d06a5a68b38ed74cd9a5741ade44c49173b6

PR #425, opened 2022-07-24
fix: renamed callSafety -> callSafely, it seems to be more correct name

whiteand:


artalar:
It is good that we not in production already %) thank you!

PR #428, opened 2022-07-24
WeakMap primitive: copied atomizeMap and replaced Map with WeakMap

whiteand:


artalar:
A weak collections don't allow to copy them, so this code will fail in real usage.
But the idea is good, thank you. May be we should made an own implementation of WeakMap by regular Map and WeakRef. Let me know if you want to get this task.

PR #430, opened 2022-08-09
Update shallowEqual util

iminside:
Hi!
It doesn't make sense to compare all properties if somewhere in the middle there are already non-matching ones.


BataevDaniil:
Possible just use 'some' and negative condition?
This condition  looks scary. 
Thanks for finding this!

iminside:
@BataevDaniil 
Yes you are right. 
This can indeed be done through `some`. 
Fixed

artalar:
I don't understand the difference:) what kind of behavior changed by this pr?

iminside:
Oops :)
I have never done a more stupid PR :)

PR #434, opened 2022-10-29
Patch 1

vkozio:


artalar:
I think we need something like [that](https://github.com/artalar/reatom/blob/v1-lts/packages/core/src/shared.ts#L52-L68) (without symbol or array support)

Btw, I could tag it as `#hacktoberfest`

vkozio:
Sorry, created this accidentally. Github is too smart sometimes) But anyway ability to use action name generator is what I'm seriously love to see) Thanks

artalar:
Do you have a plan to make PR?

vkozio:
Yes, just will do it proper way. This one was by mistake, sorry again)

PR #435, opened 2022-10-30
feat(core-v2): store options type formatter

vkozio:


artalar:
@vkozio thank you for your work, but you should be noticed that not all reatom users uses `store` option in atoms. For SSR environment store create dynamically and used by react context, so it never used as atom option.

The name creator should be external function, like [here](https://github.com/artalar/reatom/blob/v1-lts/packages/core/src/shared.ts#L62-L68)

Akiyamka:
@artalar maybe users that use default store also ok with default store settings.
also when it option of store it much easier to test

vkozio:
Maybe we can just pass atom name and action name in separate fields here? 
https://github.com/artalar/reatom/blob/95526b2233b0218df44896a367fa59f200fa583c/packages/core/src/createAtom.ts#L104-L108

```js
 const actionCreator = (...a: any[]) => ({ 
   payload: dep(...a), 
   type, // <---------- action name
   atomId, // <--------- atom id
   targets: [atom], 
 }) 
```
Not sure if "atomId" looks good, "src" may be good name here, what do you think?
And it will perfectly solve the issue with minimal changes

artalar:
The default way should works like [that](https://github.com/artalar/reatom/blob/v1-lts/packages/core/src/shared.ts#L62-L68), but you could add additional abilities. You want to use store for that? It isn't clear for me, any other sugestions?

artalar:
Outdated. Please, reopen if needed

PR #441, opened 2022-11-10
Grammarly updates

whiteand:
Inserted first two paragrapha into grammarly and applied some advices. Maybe it is still has some errors.

There is is still long sentences which are hard to read and understand, but I am not sure I know how to improve it.


artalar:
Thanks)

PR #481, opened 2023-02-26
fix #476 docs: add install section to readme

### LINKED ISSUE #476: Docs: Readme
I really need string with package installing in readme))

pivaszbs:
fix #476


artalar:
Good idea! Please, separate scripts and move `core` right before the code of simple example, move `framework` installation right before the code of advanced example.

PR #482, opened 2023-02-26
fix #475 docs: fix spyChange example

### LINKED ISSUE #475: Docs: typo on spyChange
spyChange example doesn't have `return state` statement.
Also I think it would be better if example would be fully worked))

pivaszbs:
fix #475


artalar:
thx

Issue #487, opened 2023-03-01
eslint plugin


artalar:
We need a plugin for ESlint that will much variables and properties names and suggest to fix it.

Targets: `atom(`, `action(`, `reatom*(`.

Examples.

```ts
const countAtom = atom(0);
// -->
const countAtom = atom(0, "countAtom");

const countAtom = atom(0, "count");
// -->
const countAtom = atom(0, "countAtom");
```

```ts
const doSome = action();
// -->
const doSome = action("doSome");

const doSome = action("do");
// -->
const doSome = action("doSome");

const doSome = action(() => {});
// -->
const doSome = action(() => {}, "doSome");

const doSome = action(() => {}, "do");
// -->
const doSome = action(() => {}, "doSome");
```

```ts
const fetchUser = reatomAsync(() => {});
// -->
const fetchUser = reatomAsync(() => {}, "fetchUser");

const fetchUser = reatomAsync(() => {}, "fetch");
// -->
const fetchUser = reatomAsync(() => {}, "fetchUser");

const fetchUser = reatomAsync(() => {}, { onRequest: () => {} });
// -->
const fetchUser = reatomAsync(() => {}, {
  name: "fetchUser",
  onRequest: () => {},
});

const fetchUser = reatomAsync(() => {}, { name: "fetch" });
// -->
const fetchUser = reatomAsync(() => {}, { name: "fetchUser" });
```

artalar:
Done

PR #488, opened 2023-03-01
487 add eslint

pivaszbs:
add eslint plugin #487


packages/eslint-plugin/src/rules/action-rule.ts
```diff
@@ -0,0 +1,79 @@
...
+    create: function (context: Rule.RuleContext): Rule.RuleListener {
+        return {
+            VariableDeclarator: d => {
+                if (!isActionVariableDeclarator(d)) return;
+
+                const initArgs = d.init.arguments;
+
+                if (initArgs.length === 0) {
+                    context.report({
+                        message: `action name is not defined`,
```
vovanezha: I'd change a message a bit to make it more user friendly. For example, this could be like: `action "{{ varName }}" should has a name`, where `varName` is the variable name

packages/eslint-plugin/src/rules/action-rule.ts
```diff
@@ -0,0 +1,79 @@
...
+                        node: d,
+                        fix: fixer => fixer.replaceText(d.init, `action("${d.id.name}")`)
+                    });
+                    return;
+                }
+
+                if (isLiteral(initArgs[0]) && initArgs[0].value !== d.id.name) {
+                    context.report({
+                        message: `action name is defined bad`,
```
vovanezha: this could be like: `action "{{ varName }}" should be named as it's variable name, rename it to "{{ varName }}"`

packages/eslint-plugin/src/rules/action-rule.ts
```diff
@@ -0,0 +1,79 @@
...
+type ActionVariableDeclarator = VariableDeclarator & { id: Identifier, init: ActionCallExpression }
+
+export const actionRule: Rule.RuleModule = {
+    meta: {
+        type: 'suggestion',
+        docs: {
+            description: "Add name for every action call"
+        },
+        fixable: 'code'
+    },
```
vovanezha, line 20: I'd define all possible messages here to make a code cleaner:

```suggestion
    meta: {
        type: 'suggestion',
        docs: {
            description: "Add name for every action call"
        },
        messages: {
          noname: /* message */,
          invalidName: /* message */
        },
        fixable: 'code'
    },
```

What do you think?

packages/eslint-plugin/src/rules/action-rule.ts
pivaszbs: Oh, I think it will be great!

packages/eslint-plugin/src/rules/action-rule.ts
```diff
@@ -0,0 +1,79 @@
...
+    create: function (context: Rule.RuleContext): Rule.RuleListener {
+        return {
+            VariableDeclarator: d => {
+                if (!isActionVariableDeclarator(d)) return;
```
vovanezha: Maybe we also have to check if the current declaration is imported from `@reatom`?

packages/eslint-plugin/src/rules/action-rule.ts
pivaszbs: I think it's not good idea as it's not easy to pass variables inside)

But put messages as separate functions at the top is great

packages/eslint-plugin/src/rules/action-rule.ts
pivaszbs: Maybe, but I think it's not easy task (mb I am wrong, I didn't test it on files actually)

pivaszbs:
Also I have problems with test runner @artalar it doesn't show anything at all in my local runs (throw errors if any, but clear term if no errors)
Mb it's cause of eslint testRunner. Could you help please?

packages/eslint-plugin/src/rules/action-rule.ts
vovanezha: as far as I remember, we could do it like this:

```javascript
  create: function (context) {
  let importedFromReatom = false;
  
  return {
    ImportSpecifier(node) {
        const imported = node.imported.name
        const from = node.parent.source.value;
        if (from.startsWith('@reatom') && imported === 'action') {
          importedFromReatom = true;
        }
      },
     ...
  }
```

(i didn't test this code on your pr, just took it from one of my project :))

pivaszbs: Make sense)

packages/eslint-plugin/src/rules/action-rule.ts
vovanezha: why not easy? with a structure above you can pass variables like this:

```javascript
    context.report({
                messageId: "noname",
                data: { varName: varName },
                ...
              });
```

pivaszbs: Is there any advantages on such a solution in comparison with simple function?

`message: noname(d.id.name),`

vovanezha: 1. we use builtin tools
2. all rule configuration in one place

pivaszbs: anyway it's not typed and a bit harder to read)) (for me)

artalar:
> Also I have problems with test runner @artalar it doesn't show anything at all in my local runs (throw errors if any, but clear term if no errors) Mb it's cause of eslint testRunner. Could you help please?

@pivaszbs you run `npm run test` from the root? Attach screenshot with terminal output pls.

packages/eslint-plugin/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+module.exports = {
+  "parserOptions": {
+    "ecmaVersion": "latest"
+  },
+  root: true,
+  extends: [
```
artalar, line 8: Why does it needed?

packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,29 @@
...
+```sh
+npm i @reatom/eslint-plugin
+```
+
+## Usage
+
+```ts
+// .eslintrc.js
+{
+    ...,
```
artalar:
```suggestion
    //...
```
Make it easy to copy-pase

packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,29 @@
...
+    // use all rules
+    extends: [
+        "plugin:@reatom/recommended"
+    ],
+    ...
+    // or pick some
+    rules: {
+        '@reatom/atom-rule': 'error',
+        '@reatom/action-rule': 'error',
+        '@reatom/reatom-prefux-rule': 'error'
```
artalar: ?
```suggestion
        '@reatom/reatom-prefix-rule': 'error'
```

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "exports": {
+    "require": "./build/index.js",
+    "default": "./build/index.module.js"
+  },
+  "main": "build/index.js",
+  "module": "build/index.module.js",
+  "unpkg": "build/index.umd.js",
+  "types": "build/index.d.ts",
+  "browserslist": [
+    "last 4 chrome versions"
```
artalar:
```suggestion
    "node 16"
```

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "scripts": {
+    "sandbox": "vite",
+    "prepublishOnly": "npm run build && npm run test",
+    "build": "microbundle",
+    "test": "tsx src/**/*.test.ts",
+    "test:watch": "tsx watch src/**/*.test.ts"
+  },
+  "dependencies": {
+    "@reatom/core": ">=3.1.0"
+  },
```
artalar: It is not used

```suggestion
```

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "author": "artalar",
+  "maintainers": [
```
artalar, line 27: Add yourself! =D

packages/eslint-plugin/.eslintrc.js
pivaszbs: good part is eslint rule for eslint plugins)

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/artalar/reatom.git"
+  },
+  "bugs": {
+    "url": "https://github.com/artalar/reatom/issues"
+  },
+  "homepage": "https://github.com/artalar/reatom/packages/eslint-plugin",
```
artalar:
```suggestion
  "homepage": "https://www.reatom.dev/packages/eslint-plugin",
```

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "files": [
+    "/build",
+    "/package.json"
+  ],
+  "devDependencies": {
```
artalar, line 50: Maybe we should use less strict version, from `8.0.0`? To be much compatible with current stack of a user.

pivaszbs: It's dev, not peer, but yeah, mb it's better to not use new features)

packages/eslint-plugin/src/tests/rule.test.ts
```diff
@@ -0,0 +1,241 @@
...
+tester.run('reatom/atom-rule', atomRule, {
+    valid: [
+        {
+            code: `
+            import { atom } from '@reatom/framework'
+            const countAtom = atom(0, "countAtom");
+            `
+        },
+        {
+            code: `const countAtom = atom(0);`,
```
artalar, line 22: What this test mean? We trying to not touch `atoms`'s from different imports?

pivaszbs: Exactly

packages/eslint-plugin/src/tests/rule.test.ts
```diff
@@ -0,0 +1,241 @@
...
+              `
+        },
+        {
+            code: `
+            import { reatomAsync } from '@reatom/framework'
+            const fetchUser = reatomAsync(() => {}, { name: "fetchUser" });
+            `
+        },
+        {
+            code: `const fetchUser = reatomAsync(() => {});`,
```
artalar: We should handle **all** `reatom*` methods, it doesn't metter what import it from.

pivaszbs:
> > Also I have problems with test runner @artalar it doesn't show anything at all in my local runs (throw errors if any, but clear term if no errors) Mb it's cause of eslint testRunner. Could you help please?
> 
> @pivaszbs you run `npm run test` from the root? Attach screenshot with terminal output pls.

In root
image

In package
image

packages/eslint-plugin/package.json
artalar: Got it, my mistake)

packages/eslint-plugin/src/tests/rule.test.ts
pivaszbs: Well, again, no democracy((
Ok, mb it's better

artalar:
@pivaszbs pull from the mester and retry tests.

How do you think, we are ready to merge this?

pivaszbs:
@artalar yeah we are ready :)

artalar:
@pivaszbs there are still three `pref`**`u`**`x` in the code, fix it pls :)

artalar:
Thank you!

Issue #360, opened 2021-05-27
Replace husky by simple-git-hooks


artalar:
Add:

- [x] simple-git-hooks
- [x] commitlint
- [ ] `lint-staged` for prettier

JeremyTechDev:
Heey, can I get this issue assigned to me? Want to participate in Hacktoberfest

shameondev:
Looks like it is closed for now... @artalar

artalar:
I just fixed the description, the last missed task is `lint-staged for prettier`

PR #498, opened 2023-03-09
Add lint-staged pre-commit hook with prettier

BANOnotIT:
https://github.com/artalar/reatom/issues/360

Do you still need simpe-git-hooks, or husky is ok?


romadryud:
@BANOnotIT What you think about pre push hook instead?

artalar:
@BANOnotIT @romadryud any updates?

romadryud:
@artalar I think we can close this one, we will do it in CI as we discussed.

PR #500, opened 2023-03-13
CI: Setup release-please bot

BANOnotIT:


artalar:
@BANOnotIT could you write a short guide please of how to use it regularly and what I need to do for a first release?

BANOnotIT:
Yeah, sure. Here's how it work:
1. You commit things with Conventional Commits
2. Release please looks for commits history from "previous release" (that is it's own previous commit or bootstrap commit that is configured already)
3. Looks for fix, feat/feature, breaking change commits. Then it matches these commits against packages' files.
4. If packages were affected it'll create one commit message in special branch that'll contain version bumps with changelog for all affected packages and also create a PR for that branch if it doesn't exist yet.
5. When you want to release something you just merge the PR and that commit becomes a "previous release" in step 2
6. It also manages versions with special tags that have format `{component}-{version}`. And it also create a github release whenever you merge that PR

It also uses `.release-please-manifest.json` to track previous package versions in monorepo. This is a system file but it can be used to know which packages to publish in future. But honestly current publish script is less fragile

BANOnotIT:
Hm, for the first release we should know which commit was the last published. That commit full hash should be set in `bootstrap-sha` option.

PR #515, opened 2023-03-19
feat: add web-fetch

### LINKED ISSUE #507: Create reatom/web-fetch
We need a meta package on top of `reatomAsync`, which will accepts all `fetch` parameters and a few more options to extends `Request`.

Some feature examples

```ts
const getUser = reatomAsync(() =>
  fetch("/api/user").then((r): Promise<User> => {
    throwReatomError(!r.ok, r.statusText);
    return r.headers.get("content-type") === "application/json" ? r.json() : r;
  })
);
const getUser = reatomFetch<User>("/api/user");
const getUser = reatomFetch("/api/user", { validate: User.check });
const getUser = reatomFetch({
  url: "/api/user",
  validate: User.check,
});
```

By default the first passed argument should be stringified to the `body`.

```ts
const postUser = reatomAsync((ctx, user: User) =>
  fetch({
    url: "/api/user",
    method: "POST",
    body: JSON.stringify(user),
  })
);
const postUser = reatomFetch.post<User>("/api/user");
```

For dynamic paths or the need for some mappings, it should be possible to pass the request creation function.

```ts
const getUser = reatomFetch<User, string>((ctx, id) => `/api/user/${id}`);
const getList = reatomFetch<List, string, 'price' | 'orders'>((ctx, filter, sortBy) => ({
  url: '/api/list',
  searchParams: { filter, sortBy },
}))
```

Something for inspiration: https://github.com/developit/redaxios

ghost:
Close #507 by adding `@reatom/web-fetch` package. 

For some reason, I couldn't get `@reatom/testing` package working locally, so tests are implemented with plain `uvu`, resulting in ugly sources. I'll try to fix that later.

### DIFF
--- a/packages/web-fetch/src/index.ts
+++ b/packages/web-fetch/src/index.ts
@@ -1,0 +1,97 @@
+import {AsyncAction, AsyncCtx, reatomAsync} from '@reatom/async'
+import {throwReatomError} from '@reatom/core'
+
+export interface ReatomFetchRequestOptions {
+	fetch?: (path: string, init: RequestInit) => Promise<Response>
+	basePath?: string
+	auth?: string
+	headers?: HeadersInit
+	withCredentials?: boolean
+	validateStatus?: (status: number) => boolean | string
+	transformRequest?: (body: any, headers: Headers) => any
+}
+
+export interface ReatomFetchGetOptions<T = any>
+	extends ReatomFetchRequestOptions {
+	validate?: (result: T) => boolean | string
+}
+
+const globalDefaults = {
+	fetch,
+	basePath: '',
+	withCredentials: false,
+	validateStatus: status => status >= 200 && status < 300,
+	validate: () => true,
+	transformRequest: body => body,
+} satisfies ReatomFetchGetOptions<any>
+
+export const createReatomFetch = (defaults: ReatomFetchRequestOptions) => {
+	const req = async <T>(
+		ctx: AsyncCtx,
+		method: 'get' | 'post',
+		actionPath: string,
+		actionOptions?: ReatomFetchGetOptions<T>,
+		body?: T
+	) => {
+		const options = {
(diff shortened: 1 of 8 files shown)


packages/web-fetch/src/index.ts
```diff
@@ -0,0 +1,97 @@
...
+			init
+		)
+
+		const statusValid = options.validateStatus(resp.status)
+		throwReatomError(
+			!statusValid,
+			(statusValid as string | false) || resp.statusText
+		)
+
+		if (method === 'get') {
```
BANOnotIT, line 71: This might not be application/json response. But you call resp.json() several lines below. Would you mind adding adding Content-type check here?

BANOnotIT: Also, post responses might have json response too

packages/web-fetch/src/sandbox.tsx
BANOnotIT, line 1: Do you really need these html and vite.config files

packages/web-fetch/src/index.ts
```diff
@@ -0,0 +1,97 @@
...
+
+			return json as T
+		}
+	}
+
+	return Object.assign(
+		<T>(path: string, options?: ReatomFetchGetOptions<T>) =>
+			reatomAsync(ctx => req(ctx, 'get', path, options) as Promise<T>),
+		{
+			post: <T>(path: string, options?: ReatomFetchRequestOptions) =>
```
BANOnotIT, line 88: Interface doesn't support fetch(resource) variant declared in issue:
```typescript
const getUser = reatomFetch({
  url: "/api/user",
  validate: User.check,
});
```

ghost: Is there any reason to support multiple formats? More complex APIs lead to messier typings and therefore less readable errors and language server highlights.

PR #518, opened 2023-03-28
Typed atom reducer

BANOnotIT:


BANOnotIT:
`atom((ctx, prevState = 1) => prevState + 1)` can't infer types in a normal way because of TS's inference limitations: 
https://stackoverflow.com/a/75866794/11714856

artalar:
outdated

PR #524, opened 2023-03-31
Use fontsource instead of google fonts

### LINKED ISSUE #522: Use local fonts instead of google
https://github.com/artalar/reatom/blob/2e03ec68ba00ff564967a7fac822691754206c21/docs/src/layouts/Layout.astro#L40

BANOnotIT:
Do we really need Roboto if we only use Roboto Mono?

closes #522


artalar:
Btw, could we specify special `weight`? Will it reduce the size? In the docs we use `normal`, `bold`, `600`.

PR #526, opened 2023-03-31
fix #525 docs: fix onEffect example

### LINKED ISSUE #525: docs:  Incorrect parameter order in onEffect example
There is an error in the documentation for the reatomAsync function. The order of the parameters params and promise in the onEffect hook is incorrect. 

The current example provided in the documentation is:
```ts
onEffect(ctx, promise, params) {
  listAtom(ctx, { data: [], status: 'pending' })
},
```

The correct order of the parameters should be:
```ts
onEffect(ctx, params, promise) {
  listAtom(ctx, { data: [], status: 'pending' })
},
```

Please update the documentation to reflect the correct order of the parameters to prevent confusion for users working with the reatomAsync function.

skhrv:


artalar:
Fixed in d6f51cb33a3b33dcaf5c3a65202cd2b67147d7d8

Issue #489, opened 2023-03-03
withCache does not update statusesAtom


vovanezha:
### Description:

If using `withStatusesAtom` and `withCache` together and if data was taken from cache, `statuesAtom` won't be updated and always returns `isPending: true`

### Reproduction link:

https://codesandbox.io/s/withcache-issue-forked-hnilwu

vovanezha:
@artalar hey! the issue is still there with `@reatom/framework@3.2.0` version. Moreover there are new additional issues:

1.  `pendingAtom` is always `1` if `withCache` specified
2.  `Cannot create property 'toString' on number '34'` error in console

Reproduction link — https://codesandbox.io/s/withcache-issue-forked-hnilwu

artalar:
@vovanezha hey! The 'toString' error fixed! The behaviour of `pendingAtom` is expected, as cache trying to cache calls - https://github.com/artalar/reatom/blob/8a88b020069a0e667a0f0b88dce987fea5a48a19/packages/async/src/withCache.test.ts#L114-L115

PR #534, opened 2023-04-07
fix(async): #489 error in console

matpaul:
Hello,
Fix console error described in #489, just delete it because value assigned to primitive will be lost


artalar:
Awesome, thank you!

PR #538, opened 2023-04-08
Fix(eslint-plugin): add detection with import aliases

raidenmiro:


artalar:
@raidenmiro thanks a lot!

PR #546, opened 2023-04-12
fix(eslint-plugin):  #541 not fired inside the object

### LINKED ISSUE #541: [eslint] Not fired inside the object
Here are some examples from below:

```ts
const handler = {
  draggable: action(ctx => {})
} 
```

and

```ts
class SomeService {
  private someAtom = atom({})
}
```

raidenmiro:
closes #541

- [x] - Add basic test-cases
- [x] - Add fired rules inside the object
- [x] - Improve message for `reatom*` primitives
- [x] - Fix correction when using types with `action`, example from below

from:

```ts
export const changeCurrency = action<string>()
``` 

into:
```ts
export const changeCurrency = action<string>("changeCurrency")
```


raidenmiro:
@artalar, @pivaszbs, Can you review this changes when you have some free time?

artalar:
@raidenmiro thank you a lot!

Issue #529, opened 2023-04-02
[eslint] Track variable naming


artalar:
Atoms should `Atom` end.

BANOnotIT:
Or $?

artalar:
not "$" :) We should a userspace option, is it possible?

artalar:
released

PR #552, opened 2023-04-19
Add postfix checker

pivaszbs:
Add #529 feature

### DIFF
--- a/packages/eslint-plugin/src/rules/atom-postifx-rule.ts
+++ b/packages/eslint-plugin/src/rules/atom-postifx-rule.ts
@@ -1,0 +1,101 @@
+import type { Rule } from 'eslint'
+import type { CallExpression, Identifier, Literal, Node } from 'estree'
+import {
+    extractAssignedVariable,
+  extractAssignedVariableName,
+  extractImportDeclaration,
+  isLiteral,
+  traverseBy,
+} from '../lib'
+import { isAtomCallExpression } from './atom-rule'
+
+type AtomCallExpression = CallExpression & {
+  callee: Identifier
+  arguments: [Literal] | [Literal, Literal]
+}
+
+export const atomPostfixRule: Rule.RuleModule = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Check postfix for every atom',
+    },
+    fixable: 'code',
+    messages: {
+      incorrectVariableName: `atom "{{ atomName }}" should have postfix "{{ postfix }}"`,
+    },
+  },
+  create: function (context: Rule.RuleContext): Rule.RuleListener {
+    const importedFromReatom = new Map<string, string>()
+    const postfix = context.settings.atomPostfix;
+    const badPostfix = (atomName: string) => {
+        if (!postfix) return false;
+
+        return !atomName.endsWith(postfix)
+    };
+
+    return {
+      ImportDeclaration(node) {
+        extractImportDeclaration({
+          from: node,
+          importsAlias: importedFromReatom,
+          packageName: '@reatom',
+        })
+      },
+      CallExpression: (node) => {
+        if (!isAtomCallExpression(node, importedFromReatom)) return
+
+        const matchBy = new Set([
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin/src/rules/atom-postifx-rule.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export const atomPostfixRule: Rule.RuleModule = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Check postfix for every atom',
+    },
+    fixable: 'code',
+    messages: {
+      incorrectVariableName: `atom "{{ atomName }}" should have postfix "{{ postfix }}"`,
```
BANOnotIT, line 25:
```suggestion
      incorrectVariableName: `atom "{{ atomName }}" should have postfix and be "{{ correctName }}"`,
```

packages/eslint-plugin/src/rules/atom-postifx-rule.ts
```diff
@@ -0,0 +1,97 @@
...
+        if (badPostfix(atomName)) {
+            reportIncorrectVariableName({
+            context,
+            messageId: 'incorrectVariableName',
+            source: atomIdentifier,
+            incorrectName: atomName,
+            correctName: `${atomName}${postfix}`,
+            highlightNode: atomIdentifier,
+            postfix
+          })
```
BANOnotIT: Formatting problem

packages/eslint-plugin/src/rules/atom-postifx-rule.ts
```diff
@@ -0,0 +1,97 @@
...
+  incorrectName: string
+  source: Node
+  postfix: string
+}) {
+  const { source, incorrectName, correctName, highlightNode, context, postfix } = config
+
+  context.report({
+    messageId: config.messageId,
+    node: highlightNode,
+    data: { atomName: incorrectName, postfix },
```
BANOnotIT, line 92:
```suggestion
    data: { atomName: incorrectName, correctName },
```

packages/eslint-plugin/src/rules/atom-rule.ts
```diff
@@ -22,6 +22,7 @@ export const atomRule: Rule.RuleModule = {
     messages: {
       missingName: `atom "{{ atomName }}" should has a name inside atom() call`,
       unCorrectName: `atom "{{ atomName }}" should be named as it's variable name, rename it to "{{ atomName }}"`,
+      incorrectVariableName: `atom "{{ atomName }}" should have postfix "{{ postfix }}"`,
```
BANOnotIT: should we change to `{{ correctName }}` here also?

packages/eslint-plugin/src/tests/atom-postfix-rule.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      code: `
+            import { atom } from '@reatom/framework'
+            const countAtom = atom(0, "countAtom");
+            `,
+    },
+    {
+      code: `const countAtom = atom(0);`,
+    },
+    {
+      code: 'const countAtom = atom(0, "count");',
```
BANOnotIT: shouldn't it have `countAtom` here or it's not reatom exported atom and it should not be considered?

BANOnotIT (changes requested):

packages/eslint-plugin/src/tests/atom-postfix-rule.test.ts
pivaszbs: it's variable postfix checker, not name in the call. 
This test is not suitable)
I'll delete one test

packages/eslint-plugin/src/rules/atom-postifx-rule.ts
pivaszbs: This message may be not really good if it will be
countAto -> countAtoAtom (but with postfix message it can be interpreted easier)

artalar:
Thanks!

PR #568, opened 2023-05-10
fix: new way of hooking up typings #560

### LINKED ISSUE #560: TypeScript compile option `moduleResolution` in the `bundler` value causes an error `TS7016: Could not find a declaration file for module '@reatom/*`
TypeScript compile option `moduleResolution` in the `bundler` value causes an error `TS7016: Could not find a declaration file for module...` when importing `@reatom/framework`, `@reatom/npm-react` packages.

JulianCataldo:
Related to module resolution (nodenext, bundler…),
introduced in TS5

fixes #560

### DIFF
--- a/packages/all-settled/package.json
+++ b/packages/all-settled/package.json
@@ -7,7 +7,8 @@
   "source": "src/index.ts",
   "exports": {
     "require": "./build/index.js",
-    "default": "./build/index.module.js"
+    "default": "./build/index.module.js",
+    "types": "./build/index.d.ts"
   },
   "main": "build/index.js",
   "module": "build/index.module.js",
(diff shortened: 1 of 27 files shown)


artalar:
@BANOnotIT how to force release-please to mark all packages with patch bump?

BANOnotIT:
@artalar since we don't have  [node-workspace plugin](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#node-workspace) enabled you should use [multiheader format](https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features) for every package affected. That might be tedious but that's the easiest one.

PR #624, opened 2023-09-09
fix(esling-plugin): highlight atom name, not atom declaration in case…

### LINKED ISSUE #609: Eslint plugin should highlight the name, not the atom
It is not nice that ESLint highlights the atom when the error appears only in the name argument. Is it possible to fix it - highlight only the name?

CleanShot 2023-07-28 at 22 00 49@2x

fanich37:
fix #609


artalar:
@fanich37 thank you!

PR #650, opened 2023-10-01
pendingWhileRetrying -- tests help needed

BANOnotIT:


artalar:
I was thinking about it a lot... What do you think:

```ts
const someResource = reatomResource((ctx) => fetch(`/api/list?search=${ctx.spy(searchAtom)}`))
  .pipe(withRetry({ onReject(){} })
someResource.pendingAtom.pipe(withComputed(
  (ctx, state) => state || ctx.spy(someResource.retriesAtom) > 0
))
```

artalar:
Outdated?

PR #651, opened 2023-10-04
fix(form-web): keep original element type

sqrtsanta:
Currently `input` with `type: "password"` is impossible to use with form-web. Removed nullish coalescing assignment to keep original element type


artalar:
@sqrtsanta thanks)

PR #657, opened 2023-10-07
feat(jsx): tag factory

ghost:


packages/jsx/README.md
```diff
@@ -57,10 +57,88 @@ if (import.meta.env.DEV) {
...
+
+```ts
+export function Dashboard({ user }: { user: Atom<User | null> }) {
+  const authorized = atom((ctx) => <div>Dashboard content</div>)
+  const unauthorized = atom((ctx) => <AuthRedirect />)
+  return atom((ctx) =>
+    ctx.spy(user) ? ctx.spy(authorized) : ctx.spy(unathorized),
+  )
+}
+```
```
artalar: It is not obvious why you use atoms here. And the return of the component looks cleaner (as an example) with container component. Btw, let's use `const` functions by default.

```suggestion
```tsx
export const Dashboard = ({ user }: { user: Atom<User | null> }) => (
  <section>
    <h1>Dashboard</h1>
    {atom((ctx) =>
      ctx.spy(user) ? <div>Dashboard content...</div> : <AuthModal />
    )}
  </section>
);```
```

packages/jsx/README.md
```diff
@@ -57,10 +57,88 @@ if (import.meta.env.DEV) {
...
+
+```ts
+export function Dashboard({ user }: { user: Atom<User | null> }) {
+  return match(
+    user,
+    atom((ctx) => <div>Dashboard content</div>),
+    atom((ctx) => <AuthRedirect />),
+  )
+}
+```
```
artalar:
```suggestion
```tsx
export const Dashboard = ({ user }: { user: Atom<User | null> }) => (
  <section>
    <h1>Dashboard</h1>
    {match(
      user,
      <div>Dashboard content...</div>,
      <AuthModal />
    )}
  </section>
);```
```

packages/jsx/README.md
```diff
@@ -57,10 +57,88 @@ if (import.meta.env.DEV) {
...
+  return jsxAtom
+}
+```
+
+### Usage without JSX
+
+To use this module without JSX transpilation, call the components like plain JS functions and create DOM elements with `t.*` functions.
+
+```tsx
+import { t } from '@reatom/jsx'
```
artalar: Why "t"?

packages/jsx/src/jsx.d.ts
```diff
@@ -56,51 +57,10 @@ export namespace JSX {
   //   $ServerOnly?: boolean
   // }
   type Accessor<T> = () => T
-  interface Directives {}
```
artalar: Do not remove these types; they could be used later, and I don't want to sync them with the large Solid sources again.

artalar (changes requested):

artalar:
I think this kind of helper will not be popular at all. I need more feedback from other people.

Btw, you could cp the reatom/jsx docs to the separate PR.

PR #659, opened 2023-10-09
feat(logger): devtools draft

ghost:


packages/logger/src/rld-filter-model.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export type FilterAction = (typeof FilterActions)[number]
+export const FilterActions = ['hide', ...FilterColors] as const
+
+export type Filter = {
+  code: AtomMut<string>
+  action: AtomMut<FilterAction>
+}
+
+const filterMake = (code: string, action: FilterAction) => ({
```
artalar:
```suggestion
const reatomFilter = (code: string, action: FilterAction) => ({
```

packages/logger/src/rld-filter-model.ts
```diff
@@ -0,0 +1,140 @@
...
+  fromSnapshot: (ctx, snapshot = []) =>
+    (snapshot as any[]).map((filter) => filterMake(filter.code, filter.action)),
+}
+
+export const filtersSession = reatomArray(
+  [] as Filter[],
+  'filtersSession',
+).pipe(
+  withSessionStorage({
+    key: 'rld/filtersSession',
```
artalar: Please, remove all `rld` words and replace it with something more obvious

packages/logger/src/rld-filter.ts
```diff
@@ -0,0 +1,148 @@
...
+import { match } from '@reatom/lens'
+import { buttonStyles } from './rld'
+
+export function RldFilterMenu() {
+  return atom((ctx) =>
+    t.div({
+      ...menuStyles({}),
+      children: [
+        RldFilterMenuDraft(),
+        match(ctx.get(model.filters).length, t.div(draftHrStyles({}))),
```
artalar: Did you forget the `spy`?
```suggestion
        match(ctx.spy(model.filters).length, t.div(draftHrStyles({}))),
```

You could put the computation as a function to the match method to scope reactivity to the match atom (prevent rerenders).
```suggestion
        match(ctx => ctx.spy(model.filters).length, t.div(draftHrStyles({}))),
```

packages/logger/src/rld-filter.ts
```diff
@@ -0,0 +1,148 @@
...
+  code: AtomMaybe<string>
+  setCode: (ctx: Ctx, next: string) => void
+  buttonLabel: string
+  buttonClicked: (ctx: Ctx) => void
+}) {
+  return h('div', listItemStyles({}), [
+    t.input({
+      ...filterPersistStyles({}),
+      type: 'checkbox',
+      title: 'Persist filter?',
```
artalar: I was thinking about that... All filters should be persisted by default, but the checkmark should stay and control the "active" behavior. You should be able to deactivate an existing filter easily.

packages/logger/src/rld-inspect.ts
```diff
@@ -0,0 +1,203 @@
...
+          ...viewHandleTopStyles({}),
+          children: [
+            t.div({
+              ...viewHorStyles({}),
+              children: [
+                t.div({
+                  ...viewContentStyles({}),
+                  style: `width: ${ctx.spy(width)}px; height: ${ctx.spy(
+                    height,
+                  )}px;`,
```
artalar: Why don't you use css-vars?
```suggestion
                  ...viewContentStyles({
                    width: ctx.spy(width), height: ctx.spy(height)
                  }),
```

```ts
const viewContentStyles = styled('')`
  overflow-y: scroll;
  width: ${({ width: number }) => width}px;
  height: ${({ height: number }) => height}px;
`
```

packages/logger/src/rld-filter.ts
```diff
@@ -0,0 +1,148 @@
...
+        children: [action],
+      }),
+    ),
+  })
+}
+
+const menuStyles = styled('')`
+  display: flex;
+  flex-direction: column;
+  gap: 0.25rem;
```
artalar: Do not use relative units, as the original app could have weird defaults. Use px for now.

packages/logger/src/rld.ts
```diff
@@ -0,0 +1,141 @@
...
+                  }),
+                ),
+              }),
+              atom(
+                (ctx) =>
+                  ({
+                    inspect: RldInspect(),
+                    filters: RldFilterMenu(),
+                  })[ctx.spy(currentTab)],
+              ),
```
artalar:
```suggestion
              match(currentTab)
                .is('inspect', () => RldInspect())
                .is('filters', () => RldFilterMenu())
```

packages/persist/src/index.ts
```diff
@@ -125,7 +125,7 @@ export const reatomPersist = (
         ctx: Ctx,
         state: AtomState<T>,
       ): PersistRecord => ({
-        data: toSnapshot(ctx, state),
+        data: toSnapshot({ ...ctx, spy: undefined }, state),
```
artalar, line 128: You could remove it, the bug is fixed here:https://github.com/artalar/reatom/commit/344d2ce10676f49999c8c3fb973109e1ac42c57c

artalar (changes requested):

PR #674, opened 2023-10-18
feat(jsx): support 'style' and 'styles' props

ghost:


packages/jsx/src/index.ts
```diff
@@ -42,6 +42,43 @@ export const reatomJsx = (ctx: Ctx) => {
     })
   }
 
+  const renderProp = (element: any, key: any, val: any) => {
+    if (key === 'style') {
+      for (const style in val) element.style.setProperty(style, val[style])
+    } else if (key === 'styles') {
+      element.style = val.style
```
```suggestion
      for (const style in val.style) element.style.setProperty(style, val[style])
```

packages/jsx/src/jsx.d.ts
```diff
@@ -564,6 +564,7 @@ export namespace JSX {
     lang?: AtomMaybe<string>
     spellcheck?: AtomMaybe<boolean>
     style?: AtomMaybe<CSSProperties | string>
+    styles?: AtomMaybe<{ style: string; className: string }>
```
artalar:
```suggestion
    $attrs?: Atom<Rec>
```

But please, describe the real attributes, instead of `Rec`

PR #685, opened 2023-10-22
ci: forward pull request command comment to review comment that will trigger CI

romadryud:
# Description:

This pr is aims to create a workflow which will be able to forward command comments in pr like `/prettify` to the review comments. Why do we need this forwarding? Simply because GitHub does not allow to listen `pull_request_comment` event and [force](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_comment-use-issue_comment) to use `issue_comment` event which will be triggered only when:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

So it's means that all related workflows must be in the default branch to be visible for workflow run, this makes impossible to test new CI changes in pull request if they need to be triggered on comment. Why is it impossible? Because in the note tells us that workflow will be run on the default branch, but our changes are in pr, so they just don't exist in the default branch. Changing all the time default branch also not a good way. 

On which command workflow will be triggered is defined by a variable called `COMMANDS_TO_FORWARD`, example value: ["/prettify", “/release”]

All needed instruction is provided to @artalar

### DIFF
--- a/.github/workflows/forward-command.yml
+++ b/.github/workflows/forward-command.yml
@@ -1,0 +1,42 @@
+name: Forward comment with command to review comment
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  forward:
+    if: |
+      contains(github.event.comment.html_url, '/pull/') && 
+      startsWith(github.event.comment.body, '/') &&
+      contains(fromJSON(vars.COMMANDS_TO_FORWARD), github.event.comment.body)
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    container:
+      image: node:${{vars.NODE_VERSION}}-alpine
+    steps:
+      - name: Get pull request branch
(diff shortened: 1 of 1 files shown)


.github/workflows/forward-command.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  forward:
+    if: |
+      contains(github.event.comment.html_url, '/pull/') && 
+      startsWith(github.event.comment.body, '/') &&
+      contains(fromJSON(vars.COMMANDS_TO_FORWARD), github.event.comment.body)
```
artalar, line 12:
```suggestion
     fromJSON(vars.COMMANDS_TO_FORWARD.includes(github.event.comment.body)
```
?

artalar (approved):

PR #686, opened 2023-10-22
ci: add prettier workflow and refactor others

### LINKED ISSUE #503: Setup GitHub Action to run code style tools
To save time on developer side we need to have a CI that'll run eslint/prettier inside CI on every PR and submit suggestions so that PR author can apply them.

This decreases time required to commit something.

Relates to #360 #389

romadryud:
# Description 

In this pr we have huge workflow refactoring, and it's only the start of it. I added a workflow with the name `preview` which will be the base workflow for preview all our changes, for now, `preview` has a prettier check and test workflows. In the near future `preview` will have all kinds of preview (docs, packages) and other tools like statistics and so on.

Prettier check can produce 2 variants of exit code for errors:
- `1` means there are format errors which can be fixed
- `2` means there are format errors which **can** not be fixed 

If we have `1` variant, our bot will write a comment, "Looks like you forgot to run Prettier. Hopefully, you can automatically fix it by typing `/prettify` command." After sending, this command will trigger workflow to automatically run prettier and create a commit for it, and the original comment will be deleted by the bot.

If we have `2` variant, our workflow will fail and merging to the default branch will be impossible.

Closes #503 
This depends on https://github.com/artalar/reatom/pull/685, after merging and rebase comments with command will work.


.github/workflows/format.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  format:
+    if: |
+      github.event_name == 'pull_request_review' && 
+      github.event.review.state == 'commented' && 
+      startsWith(github.event.review.body, '/') &&
+      contains(fromJSON(vars.COMMANDS_TO_FORWARD), github.event.review.body)
```
romadryud, line 13: This if only for format job, so can be simplified to 

```suggestion
      startsWith(github.event.review.body, '/prettify')
```

package.json
```diff
@@ -8,8 +8,9 @@
     "build": "turbo run build",
     "reset": "./reset.sh",
     "test": "turbo run build --filter=testing && turbo run test",
-    "prettify": "prettier --write **/*.{js,ts,md}",
-    "prettify:watch": "onchange **/*.{js,ts,md} -- prettier --write {{changed}}",
+    "prettify": "prettier --write */**/**/*.{js,ts,md}",
```
artalar, line 11: It would be safer to check only `docs/src`, `packages/**/src`, and `packages/**/README.md`.

PR #690, opened 2023-10-24
fix(jsx): correct behaviour

ghost:


packages/jsx/README.md
```diff
@@ -43,26 +45,99 @@ const onInput = action((ctx, event) =>
...
-- No SSR support
-- No SVG support
-- No keyed lists support
+This library implements a common TypeScript JSX factory that creates and configures **native** DOM elements.
+
+By default, props passed to the JSX factory are set as attributes. Add `field:` prefix to a prop name to set element fields instead of attributes. It can be used to set stuff like `HTMLInputElement..indeterminate` and `SVGPathElement..d`. There are props that treated as fields by default:
+
+- `className` (for compatibility with React libraries like [`stylerun`](https://github.com/artalar/stylerun))
+- `innerHTML`
+- `innerText`
```
artalar, line 70: Let's avoid `innerHTML` and `innerText` defaults as they are very rarely used in regular application usage. `className` bindings may be an extra thing here too; I will think about it.

packages/jsx/README.md
```diff
@@ -43,26 +45,99 @@ const onInput = action((ctx, event) =>
...
 
-Here is an example: [https://github.com/artalar/reatom-jsx/blob/main/src/App.tsx](https://github.com/artalar/reatom-jsx/blob/main/src/App.tsx)
+Because the pattern used above is somewhat verbose and excessive atoms may cause performance overhead, `@reatom/jsx` lets you to subscribe to lifecycle hooks of an atom, created by the library internally:
+
+```tsx
+const MyWidget = () => {
+  return (
+    <div
+      onConnect={(ctx) => console.log('component connected')}
+      onDisconnect={(ctx) => console.log('component disconnected')}
```
artalar, line 130: `onDisconnect` is an extra thing. Let's stick with `onConnect(() => () => cleanup())`. Also, since this property is specific to the reatom/jsx runtime, let's mark it with a special prefix like `$connect`.

**However**, the hook bindings to the returned element are very weird (poor semantics), and we should find another way.

A possible compromise solution could be the `$ref` property, similar to React, which accepts a callback with `ctx` and the connected `element` as arguments. The callback could return the cleanup function.

packages/jsx/package.json
```diff
@@ -38,7 +38,8 @@
     "@reatom/core": ">=3.5.0",
     "@reatom/lens": "^3.6.2",
     "@reatom/utils": "^3.5.0",
-    "csstype": "^3.1.2"
+    "csstype": "^3.1.2",
+    "linkedom": "^0.16.4"
```
artalar, line 42: Move it to `devDependencies`

packages/jsx/src/index.test.ts
```diff
@@ -1,10 +1,152 @@
 import { test } from 'uvu'
 import * as assert from 'uvu/assert'
+import { parseHTML } from 'linkedom'
```
artalar, line 3: I'm glad that you set up these tests. Good job! 👍

PR #709, opened 2023-11-11
feat(eslint-plugin): add async-rule & unit-naming-rule

### LINKED ISSUE #706: Add eslint rule to handle async
We need eslint rule that force to put `ctx.schedule` after each `await` in `atom` / `action` / `reatom*`

ghost:


artalar:
@krulod please, do not remove previous rules as it is a breaking change, make a fallback

artalar (approved):

Issue #708, opened 2023-11-07
Add mermaid support for docs


artalar:
We need to load mermaid.js and render diagrams for `pre` blocks which contains `marmaid` code, like there: https://www.reatom.dev/package/form/#field-validation-behavior

qweered:
https://code.juliancataldo.com/component/astro-diagram/#doc

PR #722, opened 2023-12-04
docs: support mermaid (#708)

ghost:


docs/src/components/MarkdownContent.astro
```diff
@@ -7,98 +7,37 @@ import DefaultMarkdownContent from '@astrojs/starlight/components/MarkdownConten
...
+      _mermaidApi = mermaid
     }
 
-    const copyButton = copyButtonTemplate.cloneNode(true) as HTMLButtonElement
-    copyButton.onclick = copy
-    copyButton.title = 'Copy snippet'
-    codeBlock.appendChild(copyButton)
-  }
-</script>
+    console.log(document.documentElement.dataset)
```
artalar: garbage

docs/src/components/MarkdownContent.astro
```diff
@@ -7,98 +7,37 @@ import DefaultMarkdownContent from '@astrojs/starlight/components/MarkdownConten
...
-
-      copyButton.classList.add('highlight')
-      setTimeout(() => copyButton.classList.remove('highlight'), 500)
-    }
-
-    codeBlock.onscroll = () => {
-      const scroll = codeBlock.scrollLeft
-      requestAnimationFrame(() => {
-        codeBlock.style.setProperty('--scroll', scroll + 'px')
+  let _mermaidApi: import('mermaid').Mermaid
```
artalar: Just put it into separate module and you will be able to import in as a normal singleton. Astro should allow you to import local files as a modules even from inline scripts

artalar (changes requested):

artalar (approved):

PR #761, opened 2024-02-13
Add reatomClock utility in @reatom/timer to operate with absolute time

BANOnotIT:


artalar:
@BANOnotIT please, change the API to: 

```ts
reatomClock(
  options?: { updateInterval?: number, updateStart?: 'init' | 'connection' },
  name?: string
)
```

artalar (changes requested):
@BANOnotIT please, change the API to: 

```ts
reatomClock(
  options?: { updateInterval?: number, updateStart?: 'init' | 'connection' },
  name?: string
)
```

BANOnotIT:
How `updateStart = 'init'` should wok? Why `name` should be a separate positional argument if it's inside options object in other APIs?

artalar:
@BANOnotIT 

> How updateStart = 'init' should wok?

Just like `withInit` - on first atom touch (read / update)

> Why name should be a separate positional argument

We, probably, should improve eslint to allow passing the name in the first argument. @krulod what do you think, don't it be too complex?

BANOnotIT:
What if we just also refactor reatomTimer to accept name as second argument and deprecate name in first argument?

BANOnotIT:
Do I understand correctly that 'init' just should not stop timer on disconnect?

artalar:
@BANOnotIT

> Do I understand correctly that 'init' just should not stop timer on disconnect?

yep. AND the timer starts immidiatly on the first read, even if there is now subscribtion at all.

artalar:
I still don't understand the purpose of this helper. Please, provide an example of an application that demonstrates the obvious benefits of this method.

PR #764, opened 2024-02-16
feat(eslint-plugin): enforce domain in names [#662]

### LINKED ISSUE #662: Match domain in eslint fix
The common pattern in Reatom is to create a model in a fabric function, which should be started with the 'reatom' verb. It would be super handy to handle the model name and use it in the fix suggestion.

```ts
export const reatomMyModel = () => {
  const data = atom([]);
};
// -->
export const reatomMyModel = () => {
  const dataAtom = atom([], "MyModel.dataAtom");
};
```

```ts
const reatomMyModel = action(() => {
  const data = atom([]);
});
// -->
export const reatomMyModel = action(() => {
  const dataAtom = atom([], "MyModel.dataAtom");
}, "reatomMyModel");
```

ghost:


packages/eslint-plugin/src/rules/unit-naming-rule.test.ts
```diff
@@ -40,15 +40,29 @@ tester.run('unit-naming-rule', unitNamingRule, {
...
+    },
+    {
+      code: `${ImportAtom}; function reatomSome() { const field = atom(0, 'Some.field') }`,
+      errors: [{ messageId: 'nameIncorrect' }],
+      output: `${ImportAtom}; function reatomSome() { const field = atom(0, 'reatomSome.field') }`,
+    },
+    {
+      code: `${ImportAtom}; function reatomSome({name}) { const field = atom(0, 'field'); }`,
+      errors: [{ messageId: 'nameIncorrect' }],
+      output: `${ImportAtom}; function reatomSome({name}) { const field = atom(0, \`\${name}.field\`); }`,
```
artalar, line 60: Please, add additional tests cases (and implementation, if needed) if the fabric is an action (`const reatomSome = action(...)`).

artalar (approved):

PR #773, opened 2024-03-04
Fix learning page in docs

lid0a:
1. https://www.reatom.dev/getting-started/learning/#create-atom
untitled
I think here you meant `and`, not `add`.

2. https://www.reatom.dev/getting-started/learning/#async-actions
image
`async` keyword was missed here.


artalar (approved):

artalar:
Thanks)

PR #775, opened 2024-03-06
fix(hooks): correct isInit behaviour

ghost:
Fix the utility and add some tests

[Problem report](https://discord.com/channels/1177543707524874240/1177543707994624052/1214678057634234378)

### DIFF
--- a/packages/hooks/src/index.ts
+++ b/packages/hooks/src/index.ts
@@ -197,12 +197,16 @@ export const isConnected = (ctx: Ctx, { __reatom: proto }: Atom) =>
     return !!cache && cache.subs.size + cache.listeners.size > 0
   })
 
-const initiatedAtom = atom(
-  null as any as WeakSet<AtomProto>,
-  'isInit._initiatedAtom',
+const initializations = atom(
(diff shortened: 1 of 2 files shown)


packages/hooks/src/index.ts
```diff
@@ -197,12 +197,16 @@ export const isConnected = (ctx: Ctx, { __reatom: proto }: Atom) =>
...
-  'isInit._initiatedAtom',
+const initializations = atom(
+  null! as WeakMap<AtomProto, AtomCache>,
+  'initializations',
 )
-initiatedAtom.__reatom.initState = () => new WeakSet<AtomProto>()
+initializations.__reatom.initState = () => new WeakMap()
 export const isInit = (ctx: Ctx) => {
-  const initiated = ctx.get(initiatedAtom)
-  return initiated.has(ctx.cause.proto) || !initiated.add(ctx.cause.proto)
```
artalar, line 207: @krulod there is no need to a Map, the bug in the final check, it should be `return !initiated.has(ctx.cause.proto) || (initiated.add(ctx.cause.proto), true)`

artalar (approved):

PR #782, opened 2024-04-01
fix(jsx): mount fragment as child

crashmax-dev:


packages/jsx/src/index.ts
```diff
@@ -183,7 +183,12 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
   let hf = () => {}
 
   let mount = (target: Element, child: Element) => {
-    target.appendChild(child)
+    // Fragment as child
+    if (Array.isArray(child)) {
+      target.append(...[child].flat(Infinity))
+    } else {
+      target.appendChild(child)
+    }
```
artalar: We focused on Byte's economy =D
```suggestion
    target.append(...[child].flat(Infinity))
```

artalar (approved):

PR #785, opened 2024-04-02
fix: change type of dataAtom in withDataAtom operator without initState

### LINKED ISSUE #784: TS: withDataAtom without initState
Hello, I'm using `reatom/async` in my app. I use `withDataAtom` without default value like this:

```typescript
const asyncAction = reatomAsync(...).pipe(withDataAtom(/* empty */))
```

I need to use `asyncAction.dataAtom.reset` in my code, but TS says that `dataAtom` hasn't this method. I think it's because of [this type annotation](https://github.com/artalar/reatom/blob/33a0fda91bd1ed5a384d8214baa08f0b295861f2/packages/async/src/index.ts#L213). Are there any reasons for using `AtomMut` instead of the `AsyncDataAtom`?

If I pass `undefined` (`withDataAtom(undefined)`), everything will be fine. It's the same thing for JS, I thing it's necessary to change type from `AtomMut` to `AsyncDataAtom`

I looked at the implementation, and [it always returns](https://github.com/artalar/reatom/blob/33a0fda91bd1ed5a384d8214baa08f0b295861f2/packages/async/src/index.ts#L254:L262) `AsyncDataAtom`.

gaskeo:
fix #784

### DIFF
--- a/packages/async/src/index.ts
+++ b/packages/async/src/index.ts
@@ -210,7 +210,7 @@ export const withDataAtom: {
     T extends AsyncAction & {
       dataAtom?: AsyncDataAtom<AsyncResp<T>>
     },
-  >(): Fn<[T], T & { dataAtom: AtomMut<undefined | AsyncResp<T>> }>
+  >(): Fn<[T], T & { dataAtom: AsyncDataAtom<AsyncResp<T> | undefined> }>
   <
     T extends AsyncAction & {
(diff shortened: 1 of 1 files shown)


artalar:
@gaskeo sorry for my delay, thank you for the contribution!

artalar (approved):

PR #814, opened 2024-04-14
docs: add heading anchors

lid0a:
Added anchor links to headings. I've just copied this feature from astro docs. If you think it's useless you can ignore this pull request.
[0f2b4246-7a4b-4d15-8630-0bd504429dc7.webm](https://github.com/artalar/reatom/assets/70815650/36c45cb8-f2d6-4906-bc6d-7385704bbd4d)

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -4,30 +4,8 @@ import starlight from '@astrojs/starlight'
 import rehypeExternalLinks from 'rehype-external-links'
 import rehypeAutolinkHeadings from 'rehype-autolink-headings'
 import rehypeSlug from 'rehype-slug'
-import { h } from 'hastscript'
 
 if (!process.env.VERCEL) await $`tsx sync-readme-to-pages.ts`
-
-const AnchorLinkIcon = h(
-  'span',
-  { ariaHidden: 'true', class: 'anchor-icon' },
-  h(
-    'svg',
-    { width: 16, height: 16, viewBox: '0 0 24 24' },
-    h('path', {
-      fill: 'currentcolor',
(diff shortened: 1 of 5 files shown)


docs/astro.config.mjs
```diff
@@ -2,8 +2,33 @@ import { $ } from 'zx'
 import { defineConfig } from 'astro/config'
 import starlight from '@astrojs/starlight'
 import rehypeExternalLinks from 'rehype-external-links'
+import rehypeAutolinkHeadings from 'rehype-autolink-headings'
+import rehypeSlug from 'rehype-slug'
+import { h } from 'hastscript'
+
 if (!process.env.VERCEL) await $`tsx sync-readme-to-pages.ts`
 
+const AnchorLinkIcon = h(
```
artalar: Why does `hastscript` needed? It isn't possible to use just template literals?

artalar:
@lid0a thanks for the PR, I love it! Please, try to reduce `hastscript`, it would be nice to not add extra dependencies.

docs/astro.config.mjs
lid0a: Unfortunately, it's not possible to use string here.
As far as I understand, `rehype-autolink-headings` plugin works with [hast tree format](https://github.com/syntax-tree/hast?tab=readme-ov-file#) and expects an object with [Element interface](https://github.com/syntax-tree/hast#element), so `hastscript` simplifies working with this. `hastscript` is also used in [astro docs](https://github.com/withastro/docs/blob/4bcea269f606b9aacf487f4725c2899c27e55cd3/plugins/rehype-autolink.ts#L4), that's why I used it too.
Anyway, I've created a minimal version of `h` function from `hastscript`, which just creates an object like this:

```ts
function h(tagName, properties = {}, ...children) {
  return {
    type: 'element',
    properties,
    tagName,
    children,
  }
}
```

I've checked it in my local environment and it works properly. I can remove `hastscript` from dependencies and use this code instead and make a new commit if you want.

lid0a:
@artalar I've rethought the whole thing and decided to make it simpler. I've also got rid of the svg and used a hash sign instead. Please review new commit

artalar (approved):

artalar:
@lid0a nice! I will merge it, but can you fix this somehow?
CleanShot 2024-04-17 at 14 19 47@2x

lid0a:
> @lid0a nice! I will merge it, but can you fix this somehow? CleanShot 2024-04-17 at 14 19 47@2x

Done https://github.com/artalar/reatom/pull/819

PR #822, opened 2024-04-21
fix: rename initState to computed in atom overload

qweered:
Before

After (callback is now computed, screenshots are old)

### DIFF
--- a/packages/core/src/atom.ts
+++ b/packages/core/src/atom.ts
@@ -671,7 +671,7 @@ function onCall(this: Action, cb: Fn): Unsubscribe {
   })
 }
 
-export function atom<T>(callback: (ctx: CtxSpy) => T, name?: string): Atom<T>
+export function atom<T>(computed: (ctx: CtxSpy) => T, name?: string): Atom<T>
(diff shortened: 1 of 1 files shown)


packages/core/src/atom.ts
```diff
@@ -671,7 +671,7 @@ function onCall(this: Action, cb: Fn): Unsubscribe {
   })
 }
 
-export function atom<T>(initState: (ctx: CtxSpy) => T, name?: string): Atom<T>
+export function atom<T>(callback: (ctx: CtxSpy) => T, name?: string): Atom<T>
```
BANOnotIT: here are my options:
- reducer, because the second argument is the previous state
- computer, because when we pass a function we turn an atom into a computed atom

BANOnotIT (changes requested):

packages/core/src/atom.ts
qweered: @artalar

artalar: `computed` is the right name

BANOnotIT (approved):

PR #828, opened 2024-04-28
fix: adding initial state in historyAtom

### LINKED ISSUE #827: The initial state of atom isn't added to the historyAtom
### What happened?

Hi, We found weird behavior in `@reatom/undo`. The initial state isn't added to the historyAtom without this line: 
`ctx.get(a.historyAtom)` before first changing the atom.

I wrote the test in the package with `@reatom/undo` to show the problem:

```
test('withUndo without getting historyAtom before first change', async () => {
  const a = atom(0).pipe(withUndo({ length: 5 }))
  const ctx = createTestCtx()

  /**
   * The initial state isn't added to the historyAtom without this line. WTF??
   *
   * ctx.get(a.historyAtom)
   */

  a(ctx, 1)
  assert.is(ctx.get(a), 1)
  assert.is(ctx.get(a.isUndoAtom), true)
  assert.is(ctx.get(a.isRedoAtom), false)
  assert.equal(ctx.get(a.historyAtom), [0, 1])
})
```

The test failed without `ctx.get(a.historyAtom)` before `a(ctx, 1)`. it seems to be unexpected and implicit behavior.

### Part

Package @reatom/undo

### Version

3.3.1

_No response_

### Relevant log output

_No response_

DmitriyBadeev:
fix https://github.com/artalar/reatom/issues/827

Problem in `withInit` of the historyAtom. Initial value doesn't set immediately.

I not found how to get initValue of atom without ctx for this reason I added in `__reatom ` prop initValue and set them in historyAtom

### DIFF
--- a/packages/undo/src/index.ts
+++ b/packages/undo/src/index.ts
@@ -61,11 +61,19 @@ export const withUndo =
     throwReatomError(isAction(anAtom) || !isAtom(anAtom), 'atom expected')
 
     if (!anAtom.undo) {
-      const { name, initValue } = anAtom.__reatom
+      const { name } = anAtom.__reatom
 
       const historyAtom = anAtom.historyAtom = atom<Array<AtomState<T>>>(
-        [initValue],
+        [],
         `${name}.Undo._historyAtom`,
+      )
+
+      anAtom.pipe(
+        withInit((ctx, init) => {
+          const state = init(ctx)
+          historyAtom(ctx, [state])
+          return state
+        }),
       )
 
       const positionAtom = (anAtom.positionAtom = atom(0, `${name}._position`))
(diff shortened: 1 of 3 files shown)


packages/undo/src/index.ts
```diff
@@ -61,12 +61,12 @@ export const withUndo =
     throwReatomError(isAction(anAtom) || !isAtom(anAtom), 'atom expected')
 
     if (!anAtom.undo) {
-      const { name } = anAtom.__reatom
+      const { name, initValue } = anAtom.__reatom
 
-      const historyAtom = (anAtom.historyAtom = atom<Array<AtomState<T>>>(
-        [],
+      const historyAtom = anAtom.historyAtom = atom<Array<AtomState<T>>>(
```
artalar, line 66: There are now need to modify AtomProto, just move the init logic to the data source.

```ts
      const historyAtom = (anAtom.historyAtom = atom<Array<AtomState<T>>>(
        [],
        `${name}.Undo._historyAtom`,
      ))
      anAtom.pipe(
        withInit((ctx, init) => {
          const state = init(ctx)
          historyAtom(ctx, [state])
          return state
        }),
      )
```

BANOnotIT: *There is no need to

DmitriyBadeev: Yes, indeed. Fixed it

artalar (approved):

Issue #842, opened 2024-05-19
Enhance developers documentation clearness


jarith:
The current version contains typos and might be not so clear for understanding.
It might be enhanced by fixing typos and rewritten in other words to make it easier to read.

BANOnotIT:
Seems that effort was completed successfully. Current docs are written mostly with LLMs which doesn't generate much typos on it's own.

 Should we close this one?

jarith:
Yes! This issue was related to the previous v3 documentation

PR #843, opened 2024-05-19
docs: rewrite README.md

jarith:
Updates README.md for https://github.com/artalar/reatom/issues/842


README.md
```diff
@@ -1,27 +1,30 @@
...
+  <small>We use the [atomization](/recipes/atomization) pattern to achieve [maximum](#how-performant-reatom-is) performance </small>
 - perfect **effects management**.
-  <small>Advanced [async package](https://www.reatom.dev/package/async) allows you to describe complex async flows with caching, retrying and automatic cancellation with native `await` and `AbortController`.</small>
+  <small>Advanced [async package](/package/async) allows you to describe complex async flows, including caching, retrying and automatic cancellation with native `await` and `AbortController`.</small>
 - nice **debugging** experience.
-  <small>Each atom and action update stores [immutable cause (call) stack](https://www.reatom.dev/getting-started/debugging/) which is super helpful for debugging complex async flows. To simplify this we have [logger package](https://www.reatom.dev/package/logger).</small>
+  <small>Each atom and action updates the store's [immutable cause (call) stack](/getting-started/debugging/). It helps a&nbsp;lot in debugging complex async flows. We also provide a [logger package](/package/logger) for that.</small>
```
jarith: Should we use `ctx` instead of `store` here? 🤔 
`store` might be confusing because there is no `store` in Reatom

```suggestion
  <small>Each atom and action updates the ctx's [immutable cause (call) stack](/getting-started/debugging/). It helps a&nbsp;lot in debugging complex async flows. We also provide a [logger package](/package/logger) for that.</small>
```

BANOnotIT (changes requested):
Please do not remove full links. Links become unusable if you read this README via GitHub or https://npm.im/@reatom/core

README.md
BANOnotIT: Store is a legacy term used in v2. You can change it to Context

artalar (approved):

PR #849, opened 2024-05-22
feat(lens): support circular references in parseAtoms

### LINKED ISSUE #772: lens: parseAtoms cache

ghost:


packages/lens/src/parseAtoms.test.ts
```diff
@@ -151,4 +151,18 @@ test('should parse deep structures', () => {
   ;`👍` //?
 })
 
+test('circular references', () => {
```
artalar: The test has no "circular references" 🧐
Please, write test for the caching which you added. For example, you can use for "a" and "b" the same atom with an object and check that a snapshot contains equal references in "a" and "b".

artalar (changes requested):

PR #852, opened 2024-05-25
docs: rewrite Recipes section pages

jarith:
Updates `Recipes` docs section pages for https://github.com/artalar/reatom/issues/842

### DIFF
--- a/docs/src/content/docs/recipes/atomization.md
+++ b/docs/src/content/docs/recipes/atomization.md
@@ -1,14 +1,16 @@
 ---
 title: Atomization
-description: How to doing fabrics right with Reatom
+description: How to do factories in a correct way with Reatom
 order: 1
 ---
 
-You could store your data from the backend in atoms without any mappings, but it is a good practice to wrap some of your model slices to atoms for better control and to have access to more reactive features. The rule is simple: **mutable properties could be an atom, readonly properties shout stay a primitive**.
+You could store your backend data in atoms without any mappings, but it's a good practice to wrap parts of your model in atoms for better control and access to more reactive features.
 
-In case you have a user model with editable `name` property:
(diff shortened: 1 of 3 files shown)


artalar (dismissed):

artalar:
@jarith awesome! But please, fix conflicts.

docs/src/content/docs/recipes/atomization.md
```diff
@@ -83,19 +95,32 @@ export const updateUserName = (state, idx, name) => {
...
   nameAtom(ctx, name)
 })
 ```
 
-Note that an atom is itself a getter and a setter for its state, so you typically don't need to write an `updateUserName` action and can instead mutate the name atom directly in the relevant component. This doesn't lead to "messy streams flows" like in other observable libraries, as Reatom has `cause` tracking, which allows you to inspect the reason for each update, providing an even better debugging experience than working with plain JSON data structures.
+Note that an atom is both a getter and a setter for its state, so you usually don't need to write an `updateUserName` action.
+Instead, you can directly mutate the name atom in the relevant component.
 
-Another cool feature and one of the biggest benefits of this pattern is demonstrated when you have a computed list from another list. For example, if you map a list of JSX elements, it will re-render on each property update. This issue can be fixed only with normalization, which is a more complex and less powerful technique than atomization.
+This approach avoids the "messy stream flows" common in other observable libraries.
```
artalar: This sentense should be relative to the next sentense. "messy stream flows" isn't a case in Reatom because of the cause tracking.

```md
You can directly modify the name atom in the relevant component. In most libraries, this is an anti-pattern because it's challenging to debug what and where changes were made. However, in Reatom, you have `cause` tracking, allowing you to inspect the reason for any atom change. This provides an even better debugging experience than working with plain JSON data structures.
```

artalar (approved):

Issue #860, opened 2024-06-06
Handle boolean, null, undefined and empty string values ​​in JSX like React


kasperskei:
### What happened?

I propose an enhancement in the [@reatom/jsx](https://github.com/artalar/reatom/tree/v3/packages/jsx) library to handle boolean, null, undefined, and empty string values more gracefully. Specifically, these values should not result in unwanted text content or extra nodes in the DOM.

### Code Example:
```tsx
const emptyAtom = atom('', 'empty')
const emptyValue = ''
const trueAtom = atom(true, 'true')
const trueValue = true
const falseAtom = atom(false, 'false')
const falseValue = false
const nullAtom = atom(null, 'null')
const nullValue = null
const undefinedAtom = atom(undefined, 'undefined')
const undefinedValue = undefined

const element = <div>
  {emptyAtom}
  {emptyValue}
  {trueAtom}
  {trueValue}
  {falseAtom}
  {falseValue}
  {nullAtom}
  {nullValue}
  {undefinedAtom}
  {undefinedValue}
</div>
```

### Current Behavior:
```tsx
element.childNodes.length; // 10
element.textContent; // 'truetruefalsefalsenullnullundefinedundefined'
```

### Expected Behavior:
```tsx
element.childNodes.length; // 5
element.textContent; // ''
```

### Motivation:
Improve DX. Generate the behavior of other popular UI libraries such as React and Vue.js.

### Part

Package @reatom/jsx

### Version

3.10.0

PR #861, opened 2024-06-06
feat(jsx): handle falsy like React

kasperskei:
https://github.com/artalar/reatom/issues/860

### DIFF
--- a/packages/jsx/src/index.test.tsx
+++ b/packages/jsx/src/index.test.tsx
@@ -262,4 +262,86 @@ test('linked list', () => {
   // assert.not.ok(isConnected(ctx, jsxList))
 })
 
+test('boolean as child', () => {
+  const { h, hf } = setup()
+
+  const trueAtom = atom(true, 'true')
+  const trueValue = true
+  const falseAtom = atom(false, 'false')
+  const falseValue = false
+
+  const element = <div>
+    {trueAtom}
+    {trueValue}
+    {falseAtom}
+    {falseValue}
+  </div>
+
+  assert.is(element.childNodes.length, 2)
+  assert.is(element.textContent, '')
+})
+
+test('null as child', () => {
+  const { h, hf } = setup()
+
+  const nullAtom = atom(null, 'null')
+  const nullValue = null
+
+  const element = <div>
+    {nullAtom}
+    {nullValue}
+  </div>
+
+  assert.is(element.childNodes.length, 1)
+  assert.is(element.textContent, '')
+})
+
+test('undefined as child', () => {
+  const { h, hf } = setup()
+
+  const undefinedAtom = atom(undefined, 'undefined')
+  const undefinedValue = undefined
+
+  const element = <div>
+    {undefinedAtom}
+    {undefinedValue}
+  </div>
+
+  assert.is(element.childNodes.length, 1)
+  assert.is(element.textContent, '')
+})
+
+test('empty string as child', () => {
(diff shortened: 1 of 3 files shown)


packages/jsx/src/index.test.tsx
```diff
@@ -262,4 +262,35 @@ test('linked list', () => {
   // assert.not.ok(isConnected(ctx, jsxList))
 })
 
+test('boolean, null, undefined and empty string children', () => {
```
BANOnotIT: Could you separate this tests for different types? Like booleans, null, undefined, empty string. That way we can increase granularity.

BANOnotIT (changes requested):

packages/jsx/src/index.ts
```diff
@@ -32,6 +32,8 @@ type DomApis = Pick<
   | 'DocumentFragment'
 >
 
+const isFalsy = (value: unknown): value is JSX.Falsy => typeof value === 'boolean' || value === '' || value == null
```
artalar: `typeof value === 'boolean'` isn't "falsy", as `true` is not falsy. Please, use `isSkipped`.

packages/jsx/src/jsx.d.ts
```diff
@@ -17,6 +17,8 @@ type ElementsAttributesAtomMaybe<T extends Record<keyof any, any>> = {
 }
 
 export namespace JSX {
+  type Falsy = boolean | '' | null | undefined
```
artalar: Why does it needed?

packages/jsx/src/index.ts
```diff
@@ -251,7 +253,7 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
           if (error) throw error
           unlink(element, un)
           element.appendChild(innerChild)
-        } else {
+        } else if (!isFalsy(child)) {
```
artalar: This isn't a good idea, try to render `Atom<undefined | number>`, firstly `undefiend`, then toggle to `123`, I think it will not work. Add the test for this pls.

artalar (changes requested):

packages/jsx/src/jsx.d.ts
kasperskei: It is not necessary. Removed it.

packages/jsx/src/index.ts
kasperskei: Only static values ​​go into this block of code. Static values ​​can only be updated if the component is re-rendered. I wrote a test.

packages/jsx/src/index.test.tsx
```diff
@@ -262,4 +262,86 @@ test('linked list', () => {
...
+  const emptyStringAtom = atom('', 'emptyString')
+  const emptyStringValue = ''
+
+  const element = <div>
+    {emptyStringAtom}
+    {emptyStringValue}
+  </div>
+
+  assert.is(element.childNodes.length, 1)
+  assert.is(element.textContent, '')
```
artalar, line 326: This fails

```suggestion
  assert.is(element.textContent, '')

  emptyStringAtom(ctx, '123')
  assert.is(element.childNodes.length, 1)
  assert.is(element.textContent, '123')
```

artalar: I get it, there ie no `mount`

artalar (approved):

BANOnotIT (approved):

Issue #888, opened 2024-07-23
Better auto-completion for the HTML input element


kasperskei:
### What happened?

**Expected behavior**
![Screenshot 2024-07-23 at 2 38 42 PM](https://github.com/user-attachments/assets/77a8d28d-f87d-4a43-afd9-eac02f45df70)
![Screenshot 2024-07-23 at 2 25 53 PM](https://github.com/user-attachments/assets/22355a3f-4753-4d8b-ba9d-b4155875d5dc)

### Part

Package @reatom/jsx

### Version

3.11.3

PR #889, opened 2024-07-23
refactor(jsx): better html input attributes

kasperskei:
https://github.com/artalar/reatom/issues/888


artalar (approved):
Nice! Thank you!

PR #892, opened 2024-08-05
fix(primitives): add the 'size' prop to reatomSet

radqut:
The [reatomSet](https://github.com/artalar/reatom/blob/v3/packages/primitives/src/reatomSet.ts) doesn't have the 'size' prop that the original [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/size) has.

This PR adds the missing `size` prop.


artalar (approved):
Nice, thank you!

Issue #897, opened 2024-08-12
withAbord with strategy first-in-win doesn't work


kasperskei:
### What happened?

After executing this code the console is empty:
```ts
import { createCtx, reatomAsync, sleep, withAbort } from '@reatom/framework'

const ctx = createCtx()

const func = reatomAsync(async (ctx, value) => {
  console.log(value)
  await ctx.schedule(() => sleep())
}).pipe(withAbort({ strategy: 'first-in-win' }))

func(ctx, 1)
func(ctx, 2)
```

### Part

Package @reatom/async

### Version

3.15.2

### Relevant log output

```shell
1
```

artalar:
fixed

PR #898, opened 2024-08-12
fix(async): withAbort strategy first-in-win

kasperskei:
https://github.com/artalar/reatom/issues/897

The changes solve two problems:
- The condition was always true for the `first-to-win` strategy
- Early return from a function does not allow adding an event `abort` handler

### DIFF
--- a/packages/async/src/index.ts
+++ b/packages/async/src/index.ts
@@ -383,14 +383,10 @@ export const withAbort =
               })
             }
 
-            const pending = ctx.get(anAsync.pendingAtom)
-
-            if (strategy === 'first-in-win' && pending > 1) {
-              ctx.schedule(() => {
-                promise.controller.abort(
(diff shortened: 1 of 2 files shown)


packages/async/src/index.ts
```diff
@@ -383,11 +383,14 @@ export const withAbort =
...
-              return
+            const pending = ctx.get(anAsync.pendingAtom)
+
+            if (strategy === 'first-in-win' && pending > 1) {
+              ctx.schedule(() => {
+                promise.controller.abort(
+                  toAbortError('concurrent request (first-in-win)'),
+                )
+              })
             }
```
artalar:
```suggestion
            if (strategy === 'first-in-win' && ctx.get(anAsync.pendingAtom) > 1) {
              promise.controller.abort(
                toAbortError('concurrent request (first-in-win)'),
              )
            }
```

artalar (approved):

PR #903, opened 2024-08-28
draft: move to smartbundle

XaveScor:
It is a partial port from microbundle to [smartbunde](https://github.com/XaveScor/smartbundle).

What is not finished:
- [ ] deploy package from the `./dist` directory
- [x] unpkg support: It is not needed after the personal conversation
- [ ] browserslist support (??)


artalar:
Please, try to move the devtools package first, it would be ideal as it is a developement package and has a lowest effect to users production code

PR #909, opened 2024-08-30
docs(jsx): add details to documentation

kasperskei:


packages/jsx/README.md
```diff
@@ -201,6 +221,68 @@ const anSvgElement = (
 )
 ```
 
+When there is a need to work with SVG as a string, you can use one of the following options.
```
romadryud:
```suggestion
If you need to use SVG as a string, you can choose from these options:
```
WDYT?

packages/jsx/README.md
```diff
@@ -283,3 +365,10 @@ These limitations will be fixed in the feature
 
 - No DOM-less SSR (requires a DOM API implementation like `linkedom` to be provided)
 - No keyed lists support
+- A component should have no more than one root element. If this interferes with the layout, you can wrap the parent elements in another element with the style `display: "contents"`:
```
romadryud, line 368: Note for myself: We definitely need an eslint rule for that.

packages/jsx/src/index.test.tsx
```diff
@@ -517,7 +519,33 @@ test('ref mount and unmount callbacks order', async () => {
...
+
+test('style object update', () => {
+  const { ctx, h, hf, parent, mount, window } = setup()
+
+  const styleAtom = atom({
+    top: '0',
+    right: undefined,
+    bottom: null as unknown as undefined,
+    left: '0',
+  } as JSX.CSSProperties)
```
romadryud, line 533: Why do we need these assertions? Can we do it without it?

artalar (dismissed):

artalar (approved):

PR #920, opened 2024-09-09
feat(eslint-plugin): complete unit-naming-rule (#523)

### LINKED ISSUE #523: Improve eslint matches
We should support a couple additional cases: private prefix (`_`), domain chain (`some.atomName`), object keys as the name for property atom. **The fix suggestions should respect it and use it**. Here is a list of currently invalid tests: [unit-naming-rule.test.ts in eslint-next branch](https://github.com/artalar/reatom/blob/eslint-next/packages/eslint-plugin/src/rules/unit-naming-rule.test.ts).

Here is an article about writing an ESLint plugin: https://dev.to/xavescor/eslint-plugin-what-was-missed-in-the-doc-fmg

These changes may require a huge refactoring, it would be nice to support eslint@9 as well (cc https://github.com/artalar/reatom/issues/857).

Here is a helpful tools which you can use: https://github.com/coderaiser/putout/issues/210, https://github.com/artalar/reatom/issues/523#issuecomment-2192042677.

de-jabber:
fix #523


packages/eslint-plugin/src/rules/async-rule.test.ts
```diff
@@ -9,19 +9,18 @@ const tester = new RuleTester({
...
-  ],
-  invalid: [
-    {
-      code: `${ImportReatomAsync}; const reatomSome = reatomAsync(async ctx => await someEffect())`,
-      errors: [{ messageId: 'scheduleMissing' }],
-      output: `${ImportReatomAsync}; const reatomSome = reatomAsync(async ctx => await ctx.schedule(() => someEffect()))`,
-    },
-  ],
-})
+// tester.run('async-rule', asyncRule, {
```
artalar: Why is this commented out?

packages/eslint-plugin/README.md
```diff
@@ -102,25 +100,21 @@ const reatomFood = (config: {
...
-You can also ensure that `atom` names have a prefix or a postfix through the configuration, for example:
+You can also ensure prefixes and suffixes for `atom` names through the configuration:
 
 ```ts
-{
+;({
   atomPrefix: '',
-  atomPostfix: 'Atom',
-}
+  atomSuffix: 'Atom',
```
artalar: This is unnecessary breaking change.

packages/eslint-plugin/src/rules/unit-naming-rule.test.ts
```diff
@@ -8,61 +8,80 @@ const tester = new RuleTester({
...
       options: [{ atomPrefix: '$' }],
     },
     {
-      code: `${ImportAtom}; const someAtom = atom(0, 'someAtom')`,
-      options: [{ atomPostfix: 'Atom' }],
+      code: `const someAtom = atom(0, '_someAtom')`,
+      options: [{ atomSuffix: 'Atom' }],
+    },
+    {
+      code: `function reatomSome() { const someAtom = atom(0, 'reatomSome.someAtom') }`,
```
artalar: Can you remove the "reatom" sufix please?
```suggestion
      code: `function reatomSome() { const someAtom = atom(0, 'Some.someAtom') }`,
```

packages/eslint-plugin/src/rules/unit-naming-rule.test.ts
```diff
@@ -8,61 +8,80 @@ const tester = new RuleTester({
...
+    },
+    {
+      code: `const Atoms = { someAtom: atom(0, 'Atoms.someAtom') }`,
+    },
+    {
+      code: `function reatomSome() { const Atoms = { someAtom: atom(0, 'reatomSome.Atoms.someAtom') } }`,
```
artalar: Same.
```suggestion
      code: `function reatomSome() { const Atoms = { someAtom: atom(0, 'Some.Atoms.someAtom') } }`,
```

Btw, I really enjoy this feature, it will very helpfull in a complex busines code.

packages/eslint-plugin/src/rules/unit-naming-rule.test.ts
```diff
@@ -8,61 +8,80 @@ const tester = new RuleTester({
...
+      errors: [{ message: /domain must be set to the value of/ }],
+      output: `function reatomSome({name}) { const field = atom(0, \`\${name}.field\`); }`,
     },
     {
-      code: `${ImportAtom}; function reatomSome({name}) { const field = atom(0, 'Some.field'); }`,
-      errors: [{ messageId: 'nameIncorrect' }],
-      output: `${ImportAtom}; function reatomSome({name}) { const field = atom(0, \`\${name}.field\`); }`,
+      code: `function reatomSome({name}) { const field = atom(0, 'Some.field'); }`,
+      errors: [{ message: /domain must be set to the value/ }],
+      output: `function reatomSome({name}) { const field = atom(0, \`\${name}.field\`); }`,
```
artalar, line 89: 🔥

artalar (changes requested):

packages/eslint-plugin/README.md
```diff
@@ -75,22 +79,33 @@ The name must be equal to the name of a variable or a property an entity is assi
...
 ```ts
-const reatomFood = (config: {
-  name: string
-  calories: number
-  fat: number
-  carbs: number
-  protein: number
-}) => {
+const reatomFood = (config: { name: string; calories: number; fat: number; carbs: number; protein: number }) => {
+  const { name } = config.name
```
romadryud: Why do we need a name there? It seems like it is unused, and we have an example where it is used starting on line 108.

packages/eslint-plugin/README.md
```diff
@@ -75,22 +79,33 @@ The name must be equal to the name of a variable or a property an entity is assi
...
+  const protein = atom(config.protein, `Food.protein`)
+  return { calories, fat, carbs, protein }
+}
+```
+
+If a factory function defines a parameter or a variable named `name`, names of units created in the function must be template literals that derive their domain fragments from the value of `name`:
+
+```ts
+const reatomFood = (config: { name: string; calories: number; fat: number; carbs: number; protein: number }) => {
   const { name } = config.name
```
romadryud:
```suggestion
  const { name } = config
```

WDYT about this shorthand?

packages/eslint-plugin/README.md
```diff
@@ -100,27 +115,28 @@ const reatomFood = (config: {
...
 ```ts
 const secretState = atom(0, '_secretState')
 ```
 
-You can also ensure that `atom` names have a prefix or a postfix through the configuration, for example:
+You can also ensure prefixes and suffixes for `atom` names through the configuration:
 
 ```ts
-{
+;({
```
romadryud, line 134:
```suggestion
({
```
Can we delete this semicolon?

packages/eslint-plugin/src/shared.ts
```diff
@@ -1,102 +1,49 @@
...
-  }
-
-  return result
-}
+import type * as estree from 'estree'
+
+export const reatomFactoryList = ['atom', 'action', 'reaction'] as const
+export const reatomFactoryPattern = new RegExp(`^(reatom\\w+|${reatomFactoryList.join('|')})$`)
+
+// export const createImportMap = (packagePrefix: string) => {
```
romadryud: Can we remove commented code?

packages/eslint-plugin/src/rules/unit-naming-rule.ts
```diff
@@ -1,265 +1,228 @@
...
+      atomPostfix = '',
+      domainVariable = 'name',
+    }: {
+      atomPrefix?: string
+      atomPostfix?: string
+      domainVariable?: string
+    } = context.options[0] ?? {}
+
+    const domainScopes = [] as (Domain | null)[]
+    const idScopes = [] as (estree.Identifier | null)[]
```
romadryud, line 53:
```suggestion
    const domainScopes: (Domain | null)[] = [] 
    const idScopes: (estree.Identifier | null)[] = []
```

Lets remove assertions.

packages/eslint-plugin/src/rules/unit-naming-rule.ts
```diff
@@ -1,265 +1,228 @@
...
+            : []
+        const patterns = [...node.params, ...declarators.map((d) => d.id)]
+        const declaresDomain = !!patterns.flatMap(patternNames).find((id) => id.name === domainVariable)
+
+        if (declaresDomain) domainScopes.push({ is: 'dynamic', vary: domainVariable })
+        else if ('id' in node && node.id && node.id.name.startsWith('reatom')) {
+          domainScopes.push({ is: 'static', name: node.id.name.replace('reatom', '') })
+        } else domainScopes.push(null)
+      },
+      [`:function:exit`](node: estree.Function) {
```
romadryud: `node` is unused here, can we delete it?

packages/eslint-plugin/src/rules/unit-naming-rule.ts
romadryud: Second question is why do we need `null`? With following diff I have no errors in tests.

```diff
diff --git a/packages/eslint-plugin/src/rules/unit-naming-rule.ts b/packages/eslint-plugin/src/rules/unit-naming-rule.ts
index 5b5585db..5d906fa8 100644
--- a/packages/eslint-plugin/src/rules/unit-naming-rule.ts
+++ b/packages/eslint-plugin/src/rules/unit-naming-rule.ts
@@ -49,13 +49,15 @@ export const unitNamingRule: Rule.RuleModule = {
       domainVariable?: string
     } = context.options[0] ?? {}
 
-    const domainScopes = [] as (Domain | null)[]
-    const idScopes = [] as (estree.Identifier | null)[]
+    const domainScopes: Domain[] = []
+    const idScopes: estree.Identifier[] = []
 
     return {
       ':matches(VariableDeclarator, Property)'(node: estree.VariableDeclarator | estree.Property) {
         const id = node.type === 'Property' ? node.key : node.id
-        idScopes.push(id.type === 'Identifier' ? id : null)
+        if (id.type === 'Identifier') {
+          idScopes.push(id)
+        }
       },
       ':matches(VariableDeclarator, Property):exit'() {
         idScopes.pop()
@@ -73,7 +75,7 @@ export const unitNamingRule: Rule.RuleModule = {
         if (declaresDomain) domainScopes.push({ is: 'dynamic', vary: domainVariable })
         else if ('id' in node && node.id && node.id.name.startsWith('reatom')) {
           domainScopes.push({ is: 'static', name: node.id.name.replace('reatom', '') })
-        } else domainScopes.push(null)
+        }
       },
       [`:function:exit`](node: estree.Function) {
         domainScopes.pop()

```

packages/eslint-plugin/src/rules/unit-naming-rule.ts
```diff
@@ -1,265 +1,228 @@
...
+        idScopes.pop()
+      },
+      [`:function`](node: estree.Function) {
+        const declarators =
+          node.body.type === 'BlockStatement'
+            ? node.body.body.flatMap((statement) => {
+                return statement.type === 'VariableDeclaration' ? statement.declarations : []
+              })
+            : []
+        const patterns = [...node.params, ...declarators.map((d) => d.id)]
```
romadryud, line 70:
```suggestion
        const declarations =
          node.body.type === 'BlockStatement'
            ? node.body.body.flatMap((statement) => {
                return statement.type === 'VariableDeclaration' ? statement.declarations : []
              })
            : []
        const patterns = [...node.params, ...declarations.map((d) => d.id)]

```

As we return `statement.declarations` makes more sense for me to call variable `declarations`, wdyt?

packages/eslint-plugin/README.md
```diff
@@ -60,13 +60,17 @@ Here is an example of React + TypeScript + Prettier config with Reatom.
     "prettier/prettier": "error"
   },
   "settings": {
-    "atomPostfix": "Atom"
+    "atomSuffix": "Atom"
```
artalar:
```suggestion
    "atomPostfix": "Atom"
```

packages/eslint-plugin/README.md
```diff
@@ -75,22 +79,33 @@ The name must be equal to the name of a variable or a property an entity is assi
...
-const reatomFood = (config: {
-  name: string
-  calories: number
-  fat: number
-  carbs: number
-  protein: number
-}) => {
+const reatomFood = (config: { name: string; calories: number; fat: number; carbs: number; protein: number }) => {
+  const { name } = config.name
+  const calories = atom(config.calories, `Food.calories`)
```
artalar, line 97: This example have a mistakes

packages/eslint-plugin/src/rules/unit-naming-rule.ts
de-jabber: https://github.com/estree/estree/blob/master/es5.md#variabledeclaration

packages/eslint-plugin/src/rules/unit-naming-rule.ts
de-jabber: 1. there are no problems with assertions when they are used to specify the element type of an empty array literal
2. `*:exit` listeners unconditionally pop elements from `domainScopes` / `idScopes` stacks even if their corresponding enter listeners did not add a domain/identifier scope. although tests do not capture such cases, removing null stack elements will cause any non-factory functions, destructured variables and computed-keyed properties to break the rule

packages/eslint-plugin/README.md
de-jabber: prettier adds these if `semi` option is false to prevent code like `foo\n({})` from being parsed as `foo({})`

artalar (approved):

PR #942, opened 2024-10-14
feat(jsx): normalize class attribute

kasperskei:
Class binding like [vue.js](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes).

### DIFF
--- a/packages/jsx/src/index.test.tsx
+++ b/packages/jsx/src/index.test.tsx
@@ -303,7 +303,7 @@ it('custom component', setup((ctx, h, hf, mount, parent) => {
   const Component = (props: JSX.HTMLAttributes) => <div {...props} />
 
   assert.instance(<Component />, window.HTMLElement)
-  assert.is(((<Component draggable />) as HTMLElement).draggable, true)
+  assert.is(((<Component draggable="true" />) as HTMLElement).draggable, true)
   assert.equal(((<Component>123</Component>) as HTMLElement).innerText, '123')
 }))
 
@@ -416,6 +416,65 @@ it('css property and class attribute', setup(async (ctx, h, hf, mount, parent) =
   assert.ok(ref2.dataset.reatom)
 
   assert.is(ref1.dataset.reatom, ref2.dataset.reatom)
+}))
+
+it('css custom property', setup(async (ctx, h, hf, mount, parent) => {
+  const colorAtom = atom('red' as string | undefined)
+
+  const component = (
+    <div
+      css:first-property={colorAtom}
+      css:secondProperty={colorAtom}
+    ></div>
+  )
+
+  mount(parent, component)
+  await sleep()
+
+  assert.is(component.style.getPropertyValue('--first-property'), 'red')
+  assert.is(component.style.getPropertyValue('--secondProperty'), 'red')
+
+  colorAtom(ctx, 'green')
+
+  assert.is(component.style.getPropertyValue('--first-property'), 'green')
+  assert.is(component.style.getPropertyValue('--secondProperty'), 'green')
+
+  colorAtom(ctx, undefined)
+
+  assert.is(component.style.getPropertyValue('--first-property'), '')
+  assert.is(component.style.getPropertyValue('--secondProperty'), '')
+}))
+
+it('class and className attribute', setup(async (ctx, h, hf, mount, parent) => {
+  const classAtom = atom('' as string | undefined)
+
+  const ref1 = (<div class={classAtom}></div>)
+  const ref2 = (<div className={classAtom}></div>)
+
+  const component = (
+    <div>
+      {ref1}
+      {ref2}
+    </div>
+  )
+
+  mount(parent, component)
+  await sleep()
+
+  assert.ok(ref1.hasAttribute('class'))
+  assert.ok(ref2.hasAttribute('class'))
+
+  classAtom(ctx, 'cls')
+  assert.is(ref1.className, 'cls')
+  assert.is(ref2.className, 'cls')
+  assert.ok(ref1.hasAttribute('class'))
+  assert.ok(ref2.hasAttribute('class'))
+
+  classAtom(ctx, undefined)
+  assert.is(ref1.className, '')
+  assert.is(ref2.className, '')
+  assert.ok(!ref1.hasAttribute('class'))
+  assert.ok(!ref2.hasAttribute('class'))
 }))
 
 it('ref mount and unmount callbacks order', setup(async (ctx, h, hf, mount, parent) => {
--- a/packages/jsx/src/index.ts
+++ b/packages/jsx/src/index.ts
@@ -9,6 +9,8 @@ export type FC<Props = {}> = (props: Props & { children?: JSXElement }) => JSXEl
 
 export type { JSXElement, JSX }
 
+export { type ClassNameValue, cn } from './utils'
+
 type DomApis = Pick<
   typeof window,
   'document' | 'Node' | 'Text' | 'Element' | 'MutationObserver' | 'HTMLElement' | 'DocumentFragment'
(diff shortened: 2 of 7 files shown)


packages/jsx/src/index.test.tsx
```diff
@@ -303,7 +303,7 @@ it('custom component', setup((ctx, h, hf, mount, parent) => {
   const Component = (props: JSX.HTMLAttributes) => <div {...props} />
 
   assert.instance(<Component />, window.HTMLElement)
-  assert.is(((<Component draggable />) as HTMLElement).draggable, true)
+  assert.is(((<Component draggable="true" />) as HTMLElement).draggable, true)
```
kasperskei, line 306: Attribute is not boolean. That's why the tests failed.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

packages/jsx/src/index.ts
```diff
@@ -156,46 +189,47 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
...
+      } else if (k === 'ref') {
+        ctx.schedule(() => {
+          const cleanup = prop(ctx, element)
+          if (typeof cleanup === 'function') {
+            let list = unsubscribesMap.get(element)
+            if (!list) unsubscribesMap.set(element, (list = []))
+            unlink(element, () => cleanup(ctx, element))
+          }
+        })
+      } else if (k === 'class' && containsAtom(prop)) {
```
```suggestion
      } else if ((isAtom(prop) && !prop.__reatom.isAction) || ((k === 'class' || k === 'className') && isObject(prop) && (isAtom(prop) || (prop = reatomClassName(prop))))) {
```

Also

```ts
const reatomClassName = (value: object) =>
  atom((ctx, state = '') => {
    if (state && ctx.cause.pubs.length === 0) {
      ctx.cause.proto.computer = null
    } else {
      state = ''
      for (const name in value) {
        if (isAtom(value[name]) ? ctx.spy(value[name]) : value[name]) state += name + ' '
      }
    }

    return state
  })
```

packages/jsx/src/utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    })), '')
+  })
+
+  it('handles falsy array correctly', () => {
+    assert.is(ctx.get(cn([
+      '',
+      null,
+      undefined,
+      {},
+      [],
```
Akiyamka, line 38: It's a bit inconsistent, that we treat `{}` and `[]`   
In js world they truly, and in objects of this lib - too.

packages/jsx/src/utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      c: true,
+      d: {},
+      e: [],
+      f: atom(true),
+      g: () => undefined,
+    })), 'a b c d e f g ')
+  })
+
+  it('handles deep array correctly', () => {
+    assert.is(ctx.get(cn(['a', ['b', ['c']]])), 'a b c   ')
```
Akiyamka: this extra spaces at the end - interned behavior?

packages/jsx/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (typeof value === 'string') {
+    className = value
+  } else if (typeof value === 'function') {
+    className = reatomClassName(ctx, value(ctx))
+  } else if (Array.isArray(value)) {
+    const length = value.length
+    for (let i = 0; i < length; i++) {
+      const normalized = reatomClassName(ctx, value[i])
+      if (normalized) {
+        className += normalized + ' '
```
Akiyamka:
```suggestion
        className += normalized !== '' ? normalized + ' ' : normalized
```

Akiyamka: Should fix extra spaces

packages/jsx/src/utils.test.ts
artalar: Yep, it our optimization to not trying to trim each string

packages/jsx/src/index.ts
```diff
@@ -110,12 +112,14 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
         styleId = styles[val] = random().toString()
         stylesheet.innerText += '[data-reatom="' + styleId + '"]{' + val + '}\n'
       }
-      /** @see https://www.measurethat.net/Benchmarks/Show/11819/0/dataset-vs-setattribute */
+      /** @see https://www.measurethat.net/Benchmarks/Show/11819 */
       element.setAttribute('data-reatom', styleId)
-    } else if (key === 'style' && typeof val === 'object') {
```
artalar, line 115: Wow, why you removed the style props handling? Please, restore it and add additional tests for css vars, which will ensure that we will not remove this feature again later :)

packages/jsx/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  | undefined
+  | Array<ClassValue>
+  | Atom<ClassValue>
+  | Record<string, AtomMaybe<string | number | boolean | null | undefined | object>>
+  | ((ctx: CtxSpy) => ClassValue)
+>
+
+export const cn = (value: ClassValue): Atom<string> => atom((ctx) => reatomClassName(ctx, value))
+
+const reatomClassName = (ctx: CtxSpy, value: ClassValue): string => {
```
artalar: A thing name starts with "reatom*" if it returns an atom or an action. Name it `parseClasses`, maube, or something like that

packages/jsx/src/utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('handles complex correctly', () => {
+    assert.is(ctx.get(cn(() => atom(() => [
+      'a',
+      {b: atom(true)},
+      ['c'],
+      atom('d'),
+      () => 'e',
+    ]))), 'a b  c  d e ')
+  })
+})
```
artalar, line 86: Please, test the change behaviour of created atom to ensure that the created atom correctly subscribes (spy) to nested atom. You can change one of the existing tests

packages/jsx/src/utils.ts
```diff
@@ -0,0 +1,43 @@
+import { atom, Atom, AtomMaybe, CtxSpy, isAtom } from '@reatom/core'
+import { isObject} from '@reatom/utils'
+
+type ClassValue = AtomMaybe<
```
artalar:
```suggestion
export type ClassNameValue = AtomMaybe<
```

packages/jsx/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+const reatomClassName = (ctx: CtxSpy, value: ClassValue): string => {
+  let className = ''
+  while (isAtom(value)) value = ctx.spy(value)
+  if (typeof value === 'string') {
+    className = value
+  } else if (typeof value === 'function') {
+    className = reatomClassName(ctx, value(ctx))
+  } else if (Array.isArray(value)) {
+    const length = value.length
+    for (let i = 0; i < length; i++) {
```
artalar:
```suggestion
    for (let i = 0; i < value.length; i++) {
```

packages/jsx/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+      const normalized = reatomClassName(ctx, value[i])
+      if (normalized) {
```
artalar: Always use strict comparison in this kind utilities to improve performance
```suggestion
      if (normalized !== '') {
```

artalar (changes requested):

packages/jsx/src/index.ts
```diff
@@ -125,8 +127,15 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
...
       if (val == null || val === false) element.removeAttribute(key)
-      else if (val === true) element.setAttribute(key, '')
-      else element.setAttribute(key, String(val))
+      else {
+        val = val === true ? '' : String(val)
+        /**
+         * @see https://measurethat.net/Benchmarks/Show/54
+         * @see https://measurethat.net/Benchmarks/Show/31249
+         */
+        if (key === 'class' && element instanceof HTMLElement) element.className = val
```
artalar, line 137: Should we handle className too? Btw, don't forget to change the types
```suggestion
        if ((key === 'class' || key === 'className') && element instanceof HTMLElement) element.className = val
```

artalar (approved):

Issue #967, opened 2024-11-05
TypeError: Cannot read properties of undefined (reading ‘type’) in "@reatom/unit-naming-rule" Eslint


skhrv:
### What happened?

A TypeError occurs when ESLint attempts to lint a TypeScript file (close-button.tsx). The error is related to the @reatom/unit-naming-rule, specifically on line 24, and seems to be triggered by the usage of the Rest Operator.

	1.	Use the Rest Operator in a TypeScript file.
	2.	Run ESLint (version 9.13.0) on the file.
	3.	Observe the error.

Expected Behavior:
ESLint should correctly parse and lint the file without throwing a TypeError.

Actual Behavior:

A TypeError is thrown during linting:
Oops! Something went wrong! :(
```
ESLint: 9.13.0

TypeError: Cannot read properties of undefined (reading 'type')
Occurred while linting /src/shared/ui/button/close-button.tsx:24
Rule: "@reatom/unit-naming-rule"
    at file:///node_modules/@reatom/eslint-plugin/build/index.mjs:1:251
  ```

### Part

Package @reatom/eslint-plugin

### Version

experimental

PR #968, opened 2024-11-05
fix(eslint-plugin): typeError in reatom/unit-naming-rule

skhrv:
https://github.com/artalar/reatom/issues/967
- Added estree as a dependency to ensure correct type handling and output in ESLint.
- Fixed existing issues related to type errors that were previously causing linting to fail.


artalar (approved):
It would be nice to add a tests for the failure cases. Otherwise, thanks a lot!!

Issue #998, opened 2025-01-06
Props are set as properties, not attributes


kasperskei:
### What happened?

> By default, props passed to the JSX factory are set as properties. Add attr: prefix to the name to set element attribute instead of property. 
https://www.reatom.dev/package/jsx/#reference

### TODO

- Remove undocumented namespace `prop:`;
- Props are set as properties, not attributes;
- Update types to match Element typing;
- Update documentation.

### Example

Source code:
```tsx
<input disabled/>
```

Expected result:
```ts
el.disabled = true
```

Actual result:
```ts
el.setAttribute('disabled', '')
```

### Part

Package @reatom/jsx

### Version

3.16.0

PR #999, opened 2025-01-06
fix(jsx): props are set as properties

kasperskei:
https://github.com/artalar/reatom/issues/998

Changes the approach from `attribute first` to `property first`. If there is a property with this name, then we set the value to the property, otherwise we set the value to the attribute. If the property exists, but we want to set the attribute, then `attr:` is added to the property name, for example, `attr:value="abc"`

### DIFF
--- a/packages/jsx/src/index.test.tsx
+++ b/packages/jsx/src/index.test.tsx
@@ -871,3 +871,148 @@ test('linked list', () =>
       '<div><!----><span>1</span><!--test--><br><!--test--><!----></div>',
     )
   }))
+
+const expectHtmlElementProperty = <
+  Tag extends keyof JSX.HTMLElementTags,
+  Property extends keyof JSX.HTMLElementTags[Tag],
+  Value extends JSX.HTMLElementTags[Tag][Property],
+  Expected extends Tag extends keyof HTMLElementTagNameMap
+    ? Property extends keyof HTMLElementTagNameMap[Tag]
+      ? HTMLElementTagNameMap[Tag][Property]
+      : never
+    : never,
+>(
+  tag: Tag,
+  prop: Property & string,
+  value: Value,
+  expected: Expected,
+  hasAttr: boolean,
+  getAttr: null | string,
+) => {
+  const element = h(tag, { [prop]: value }) as HTMLElement
+  expect((element as any)[prop]).toBe(expected)
+  expect(element.hasAttribute(prop.toLowerCase())).toBe(hasAttr)
+  expect(element.getAttribute(prop.toLowerCase())).toBe(getAttr)
+}
+
+test('width property', () =>
+  root.start(async () => {
+    expectHtmlElementProperty('img', 'width', undefined, 0, false, null)
+    // @ts-expect-error TODO fix types
+    expectHtmlElementProperty('img', 'width', null, 0, false, null)
+    expectHtmlElementProperty('img', 'width', 1, 1, true, '1')
+    expectHtmlElementProperty('img', 'width', '1', 1, true, '1')
+    expectHtmlElementProperty('img', 'width', -1, 0, true, '-1')
+  }))
+
+test('height property', () =>
+  root.start(async () => {
+    expectHtmlElementProperty('img', 'height', undefined, 0, false, null)
+    // @ts-expect-error TODO fix types
+    expectHtmlElementProperty('img', 'height', null, 0, false, null)
+    expectHtmlElementProperty('img', 'height', 1, 1, true, '1')
+    expectHtmlElementProperty('img', 'height', '1', 1, true, '1')
+    expectHtmlElementProperty('img', 'height', -1, 0, true, '-1')
+  }))
+
+test('download property', () =>
+  root.start(async () => {
+    expectHtmlElementProperty('a', 'download', undefined, '', false, null)
+    // @ts-expect-error TODO fix types
+    expectHtmlElementProperty('a', 'download', null, '', false, null)
+    // expectHtmlElementProperty('a', 'download', false, '', false, null)
+    // expectHtmlElementProperty('a', 'download', true, '', true, '')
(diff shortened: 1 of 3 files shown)


packages/jsx/src/index.ts
```diff
@@ -84,11 +84,13 @@ const walkLinkedList = (ctx: Ctx, el: JSX.Element, list: Atom<LinkedList<LLNode<
...
 
 export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
-  const StylesheetId = 'reatom-jsx-styles'
   let styles: Rec<string> = {}
-  let stylesheet: HTMLStyleElement | undefined
+  let stylesheet: HTMLStyleElement = DOM.document.createElement('style')
+  stylesheet.id = 'reatom-jsx-styles'
   let name = ''
 
+  DOM.document.head.appendChild(stylesheet)
```
artalar: If a user will not using css prop there will unnecessary style element in their DOM.

packages/jsx/src/index.test.tsx
```diff
@@ -445,38 +439,6 @@ it('css custom property', setup(async (ctx, h, hf, mount, parent) => {
   assert.is(component.style.getPropertyValue('--secondProperty'), '')
 }))
 
-it('class and className attribute', setup(async (ctx, h, hf, mount, parent) => {
```
artalar: We need to stay a special case for `class` and `className`, it should work without prefixes in both ways.

artalar (changes requested):
LGFM, there only need to do the types changes

packages/jsx/src/index.test.tsx
kasperskei: I suggest keeping only `className`:
- Supporting only one way to assign classes makes the library's API simpler. If two options (class and className) are available, it can cause confusion and increase the likelihood of errors, such as attempting to use both properties simultaneously.
- Simplification of internal logic. It eliminates the need to decide which property takes precedence in case of a conflict.
- Most developers working with JSX are already accustomed to using className. Adding support for class may introduce unnecessary inconsistencies and complicate transitions between projects.
- Including class increases mental overhead by introducing an exception. It is much easier to understand that a property in JSX corresponds directly to an element's property.

packages/jsx/src/index.ts
kasperskei: I made this change because dynamically adding a style tag can introduce style order. When changing the order, random bugs with styles can occur.

artalar: Agree. Please, add an option to the fabric to get an ability to disable the element insertion.

packages/jsx/src/index.test.tsx
artalar: There are a valid points, but it are all controversial.
- there is no big API overhead for only this special case
- many developers which didn't know JSX specific may be confused, there are a lot of templates engines (including Preact!) which support "class"
- I think most part of developers doesn't know \ understand \ remember a difference between attributes and properties.

And also, the change that you want to do it a breaking change and a new major release - it's not worth it

packages/jsx/src/index.ts
kasperskei: Replaced Record with Map, because Map is more productive when dynamically adding and reading.

packages/jsx/src/index.test.tsx
kasperskei: I looked at the preact source code, came across the exception handling and moved it to reatom/jsx
https://github.com/preactjs/preact/blob/main/src/diff/children.js#L101

packages/jsx/src/index.ts
```diff
@@ -83,12 +98,21 @@ const walkLinkedList = (ctx: Ctx, el: JSX.Element, list: Atom<LinkedList<LLNode<
...
+export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window, options?: {
+  /**
+   * Adds a style element containing styles from the `css` property to the document.
+   * @default true
+   */
+  appendStylesheet?: boolean
+}) => {
   let name = ''
 
+  /** @see https://www.measurethat.net/Benchmarks/Show/11290 */
```
artalar: Please, measure relative **string** values
https://www.measurethat.net/Benchmarks/Show/33272/2/map-vs-object-real-world-performance-with-string-keys

In desktop FF and mobile chrome there is no real difference. But I **think** the Map could took more memory, so I purpose to stay with an object.

packages/jsx/src/index.ts
```diff
@@ -16,6 +16,21 @@ type DomApis = Pick<
   'document' | 'Node' | 'Text' | 'Element' | 'MutationObserver' | 'HTMLElement' | 'DocumentFragment'
 >
 
+const propertiesAsAttribute = [
```
```suggestion
// https://github.com/preactjs/preact/blob/d16a34e275e31afd6738a9f82b5ba2fb9dbf032b/src/diff/props.js#L107
const propertiesAsAttribute = [
```

packages/jsx/src/index.ts
```diff
@@ -119,24 +136,28 @@ export const reatomJsx = (ctx: Ctx, DOM: DomApis = globalThis.window) => {
         if (val[key] == null) element.style.removeProperty(key)
         else element.style.setProperty(key, val[key])
       }
-    } else if (key.startsWith('prop:')) {
-      // @ts-expect-error
-      element[key.slice(5)] = val
+    } else if (
+      !propertiesAsAttribute.includes(key)
```
artalar: Should we use Set or object with `key in ` check instead?

artalar (changes requested):

artalar:
@kasperskei any news?

kasperskei:
> @kasperskei any news?

In terms of functionality, everything is ready, so you can run tests and merge. There is still a lack of typing for the new properties, but I was thinking of adding them in the next PR.

artalar (approved):

PR #1023, opened 2025-01-26
docs(jsx): component documentation

kasperskei:


packages/jsx/README.md
```diff
@@ -233,13 +233,83 @@ export const HeaderInput = ({ size = 0 }) => {
...
+<input
+  class="HeaderInput_123456"
+  style="--size: 0;"
+/>
+```
+
 ### Components
 
-Components in `@reatom/jsx` are just functions returning JSX elements. They neither have state nor any lifecycle associated with them. Because component instantiation boils down into function calls, features like `$spread` are not supported in them.
+Components in `@reatom/jsx` are simple functions that return JSX elements. They have no state or lifecycle and don't support features like `$spread` since they are just function calls. This makes them lightweight and efficient during execution.
```
artalar:
```suggestion
Components in `@reatom/jsx` are simple functions that return native DOM elements. They have no state or lifecycle and don't support features like `$spread` since they are just function calls. This makes them lightweight and efficient during execution.
```

packages/jsx/README.md
```diff
@@ -233,13 +233,83 @@ export const HeaderInput = ({ size = 0 }) => {
...
+const add = action((ctx) => list(ctx, (state) => [
+  ...state,
+  <li>{state.length + 1}</li>,
+]))
+
+<div>
+  <button on:click={add}>Add</button>
+  <ul>
+    {atom((ctx) => <>{...ctx.spy(list)}</>)}
+  </ul>
```
artalar: This is tooo unusual =D
```suggestion
    {atom((ctx) => <ul>{ctx.spy(list)}</ul>)}
```

packages/jsx/README.md
```diff
@@ -233,13 +233,83 @@ export const HeaderInput = ({ size = 0 }) => {
...
+</div>
+```
+
+**Do not create elements outside of components and do not reuse them.** This ensures that each element has a unique lifecycle, works correctly with atoms, and all copies are displayed in the DOM. Subscriptions to atoms are removed in the next tick after the element is unmounted from the DOM. If the element returns to the DOM, the subscriptions are not recreated.
```
artalar:
```suggestion
**Do not create elements outside of components and do not reuse them.** All created by JSX elements should be immidiatly used in the DOM as children or `mount` methos. This ensures that each element has a unique lifecycle, works correctly with atoms, and all copies are displayed in the DOM. Subscriptions to atoms are removed in the next tick after the element is unmounted from the DOM. If the element returns to the DOM, the subscriptions are not recreated.
```

artalar (changes requested):

Issue #896, opened 2024-08-10
The case with the type ZodEffects has not been handled


baddyes:
### What happened?

I need to validate the `Confirmation` field based on the input from the `NewPassword` field using the `refine` method from the `zod` package. This method returns a `ZodEffects` type, but it hasn't been handled in `reatomZod`.

https://stackblitz.com/edit/vitejs-vite-qt4qxi?file=src%2Fmain.ts

### Part

Package @reatom/npm-zod

### Version

3.10.3

### Relevant log output

```shell
TypeError: Unsupported Zod type: ZodEffects
```

Xelson:
Should be fixed in https://github.com/artalar/reatom/pull/1029

PR #1029, opened 2025-02-01
feat(npm-zod): add vitest, add support for catches, brands, pipelines, effects and etc

### LINKED ISSUE #1006: `reatomZod` generating objects with inappropriate types
### What happened?

primitives are handled incorrectly for union types
image
image
perhaps we should use `AtomMut` for union types, and the corresponding `reatom*` primitives for strict types like `number`, `boolean`, etc

### Part

Package @reatom/npm-zod

### Version

3.10.3

Firefox, Chrome, Safari, Microsoft Edge

### Relevant log output

_No response_

Xelson:
* Closes https://github.com/artalar/reatom/issues/1006, https://github.com/artalar/reatom/issues/896
* Migrated from `uvu` to `vitest`, added more tests
* Added support for: `.catch`, `.brand`, `.pipeline`, `.lazy` and all effects like `.refine`, `transform`, etc


packages/npm-zod/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@reatom/npm-zod",
+  "type": "module",
```
artalar: Unfortunately, we can't do that.

packages/npm-zod/src/index.test-d.ts
```diff
@@ -0,0 +1,102 @@
+import { expectTypeOf, test } from "vitest";
```
artalar: Lets call the file `types.test.ts`

packages/npm-zod/src/index.test-d.ts
```diff
@@ -0,0 +1,102 @@
...
+			.transform(v => v.length > 3 ? 1337 : v)
+	})
+
+	const model = reatomZod(schema);
+	type InferedSchema = z.infer<typeof schema>;
+
+	expectTypeOf(model.refine).toEqualTypeOf<AtomMut<InferedSchema['refine']>>()
+	expectTypeOf(model.transform).toEqualTypeOf<AtomMut<InferedSchema['transform']>>()
+	expectTypeOf(model.transformedRefine).toEqualTypeOf<AtomMut<InferedSchema['transformedRefine']>>()
+	expectTypeOf(model.refinedTransform).toEqualTypeOf<AtomMut<InferedSchema['refinedTransform']>>()
```
artalar: It would be better not to use `InferredSchema` as the expected type, since we are trying to test it. The expected value should be described manually, at least partially.

packages/npm-zod/src/index.test.ts
```diff
@@ -1,30 +1,140 @@
...
-  assert.is(model.readonly, "foo");
-  assert.is(ctx.get(model.n), 42);
+  expect(ctx.get(model.refine)).toBe('string')
+  expect(ctx.get(model.transform)).toBe(1337)
+})
 
-  model.s(ctx, "bar");
-  assert.equal(track.lastInput(), { n: 42, s: "bar", readonly: "foo" });
-});
+test("right values for catch", async () => {
```
artalar, line 54: Add "catch" relevant test, something like:

```ts
model.catch(ctx, ['lalal'])
expect(ctx.get(model.catch)).toBe('catch value')
```

artalar (changes requested):
Nice work! Please, fix the comments.

packages/npm-zod/src/index.test-d.ts
Xelson: > It would be better not to use `InferredSchema` as the expected type, since we are trying to test it

but we're testing infering the types of `reatomZod` to match the output of zod inference here. Can't we trust zod and use it as a standard for type inference?

packages/npm-zod/src/index.test.ts
Xelson: > model.catch(ctx, ['lalal'])

this will not be possible because the types do not allow you to set an invalid type, but I added setting the value from `catch` if parsing failed.

PR #1031, opened 2025-02-04
feat(persist-web-storage): add init() to withIndexedDb

### LINKED ISSUE #753: @reatom/persist-web-storage garbage collection and asynchronous persistence initialization
We want to:

1) Collect obsolete persistence as soon as possible, which is currently impossible because accessing a persisted record can only be done with the relative atom.
2) Initialize asynchronous storage (IndexedDB) at application startup.

Main idea:

## Persisted Atoms Registry

- It would be nice to have a separate registry with metadata about persisted atoms, so that we can manage them based on their lifetime (delete if it has expired). 
- This would be an atom with a `PersistRecord` array that would also be persisted.
- We will have some `init` method (or maybe just on app init) that does the following:
  - Loads the registry into memory from the persistence layer
  - Checks all the entries in the registry:
    - if the entry has expired - it should be deleted (gargabe collector).
    - if the entry has not expired - it should be loaded into the memory cache.
  - We can store the `loaded` status for each individual persisted atom, or maybe for the whole registry, this is negotiable.

As a result, we can do something like this when initializing the application.

```ts
await withIndexedDb.init(ctx)
// or
await withAsyncLocalStorage.init(ctx)
```

This way, persisted atoms are **guaranteed** to be initialized during rendering.

osovv:
closes #753

### DIFF
--- a/docs/src/content/docs/package/persist-web-storage.md
+++ b/docs/src/content/docs/package/persist-web-storage.md
@@ -47,9 +47,7 @@ The main goal of this adapter is to synchronize atom between tabs without pollut
 import { atom } from '@reatom/framework'
 import { withBroadcastChannel } from '@reatom/persist-web-storage'
 
-export const isAuthedAtom = atom('', 'isAuthedAtom').pipe(
-  withBroadcastChannel('isAuthedAtom'),
-)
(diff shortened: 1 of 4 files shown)


osovv:
I realized that we need to add `async init()` to `reatomPersist` with an empty default implementation. 

I'll convert PR to draft until I refactor it

artalar:
@osovv any news?

PR #1037, opened 2025-02-09
fix(lens): migrate to vitest, fix `ParseAtoms` type

Xelson:


qweered:
remove 👍; leftovers too, they are from very old testing infrastructure

PR #1038, opened 2025-02-09
feat(form): add centralized fields initialization, add support for array fields, add `withField`

Xelson:
* Added `withField` operator which uses an atom as a field
* Added centralized initialization of fields in form
* Added field arrays with nesting support (using `reatomLinkedList` underhood)

```ts
const form = reatomForm(name => ({
  user: {
    name: '', // empty string field
    email: {  // string field by passing field options explicitly
      initState: '',
      contract: (input: string) => invariant(input.includes('@'))
    },
    password: reatomField('', `${name}.user.password`),
  },

  // custom atoms also supported by piping `withField`
  subject: reatomEnum(['general', 'support'], `${name}.subject`).pipe(withField('general')) 

  // fields array with default value
  addresses: [ 
    { country: '', street: '', city: '' }
  ],
}));
```

Example with nested field arrays:
```ts
const form = reatomForm({
  addresses: fieldArray({
    initState: [
      {
        country: '',
        street: '',
        city: '',
        tags: ['defaultTag', 'defaultTag2'],
        phoneNumbers: Array<{ number: string, priority: boolean }>(),
      }
    ],
    create: (ctx, { city, country, street, tags, phoneNumbers }) => ({
      country,
      street,
      city,

      // use fieldArray overload with default factory
      tags,

      // OR define the factory explicitly to take control of the creation behavior
      phoneNumbers: fieldArray({
        initState: phoneNumbers,
        create: (ctx, { number, priority }, name) => ({
          number,
          priority: reatomBoolean(priority, name).pipe(withField(priority))
        })
      }),
    })
  }),
});
```
Here we using `create` factory to define how we want to create the actual field.

### DIFF
--- a/packages/form/sandbox/index.tsx
+++ b/packages/form/sandbox/index.tsx
@@ -1,0 +1,264 @@
+import { AtomState, createCtx } from "@reatom/core";
+import { connectLogger } from "@reatom/logger"
+import { reatomContext, useAction, useAtom } from "@reatom/npm-react"
+import { FieldAtom, FormFieldArrayAtom, reatomForm, withField } from "../src";
+import { LinkedListAtom, LinkedListLikeAtom, reatomBoolean } from "@reatom/primitives";
+import { PropsWithChildren } from 'react';
+import { createRoot } from "react-dom/client";
+import { useFormField } from "./use-form-field";
+
+const ctx = createCtx();
+connectLogger(ctx);
+
+function App() {
+	return (
+		<reatomContext.Provider value={ctx}>
+			<Form />
+		</reatomContext.Provider>
+	)
+}
+
+const root = createRoot(document.getElementById('root')!)
+root.render(<App />)
+
+const form = reatomForm(fieldArray => ({
+	username: 'vlad',
+	addresses: [
+		{
+			country: 'my country',
+			street: 'my street',
+			city: 'my city',
+			tags: ['defaultTag', 'defaultTag2'],
+			phoneNumbers: fieldArray({
+				initState: Array<{ number: string, priority: boolean }>(),
+				create: (ctx, { number, priority }) => ({
+					number,
+					priority: reatomBoolean(priority).pipe(withField(priority))
+				})
+			})
+		}
+	],
+}));
+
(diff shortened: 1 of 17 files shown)


packages/form/sandbox/index.tsx
```diff
@@ -0,0 +1,264 @@
...
+			street: 'my street',
+			city: 'my city',
+			tags: ['defaultTag', 'defaultTag2'],
+			phoneNumbers: fieldArray({
+				initState: Array<{ number: string, priority: boolean }>(),
+				create: (ctx, { number, priority }) => ({
+					number,
+					priority: reatomBoolean(priority).pipe(withField(priority))
+				})
+			})
```
artalar: I didn't agree with "fieldArray" concept. It should be just `phoneNumbers: Array<{ number: string, priority: boolean }>()` which we use to create LinkedList WITH our create function which accept and parse JSON just like the reatomForm init state. And if a user need a very custom logic they should write `phoneNumbers: reatomLinkedList(...)` or `phoneNumbers: reatomSet(...)` and so on.

artalar: Ok, lets do it as a separate exported method and will play with that

Xelson:
we have some critical issues at this stage:
~1) to start schema validation , you need to subscribe to `fieldsState`. Probably we should check schema validation for each field change, not `fieldsState` at all)~
~2) global form options do not fall into already created `reatomField`/atoms with `withField`. Maybe we should probably enclose all the options in a reactive container in order to be able to change these options for already created atoms~
3) we don't need to check the compatiblity of the `fieldsList` structure and the standard schema parsing result, otherwise, we will not be able to benefit from the validation of parsing errors (because types will prohibit using field values that do not match the schema type)
4) even if you ignore the field values and check only the key structure, the complexity of checking the compatibility of the schema and `fieldsState` remains too high, which almost always leads to `type instantiation is excessively deep and possibly infinite` error in IDE

_upd: 3, 4 points are no longer revelant because i just removed checking for types compatibility between fields tree and schema output. Looking for better approach to introduce type-safety and do not lose parsing benefits here_

PR #1050, opened 2025-02-18
feat: optimize regexp pattern

Jamshoot66:
Pattern `.*${SOME_STRING}.*` work same as `${SOME_STRING}` now, because now we only checks only is there at least one match with `RegExp.test()` and don't need groups, indexes and so on, but it need a way more steps to execute.


artalar (commented):
Nice, thanks

artalar (approved):

PR #1062, opened 2025-03-07
chore(devtools): build with vite

ghost:


packages/devtools/src/ObservableHQ.tsx
```diff
@@ -3,22 +3,34 @@ import { ctx, FC, h, mount, JSX, css, ROOT } from './jsx'
 
 // @ts-expect-error TODO write types
 import { Inspector } from '@observablehq/inspector'
-// @ts-expect-error
-import observablehqStyles from '../../../node_modules/@observablehq/inspector/dist/inspector.css'
+import observablehqStyles from '../../../node_modules/@observablehq/inspector/dist/inspector.css?raw'
```
artalar, line 6: It was a hack for Bun
```suggestion
import observablehqStyles from '@observablehq/inspector/dist/inspector.css?raw'
```

ghost: vite doesn't allow this as well. both packages use export subpath exports, so we can't just import from them any files we want, and there are no subpaths for the styles provided

artalar (approved):

PR #1063, opened 2025-03-07
feat(jsx): improvements

ghost:


packages/jsx/src/jsx-runtime.d.ts
```diff
@@ -0,0 +1 @@
+jsx.d.ts
```
artalar, line 1: ?

packages/jsx/src/jsx.d.ts
```diff
@@ -20,19 +20,9 @@ export namespace JSX {
...
   /** @todo Try replacing `Node | Element` with `ChildNode`. */
-  type ElementPrimitiveChildren =
-    | Node
-    | Element
-    | (string & {})
-    | number
-    | boolean
-    | null
-    | undefined
+  type Child = ChildNode | string | number | boolean | null | undefined
```
artalar, line 23: Do not broken change namings

packages/jsx/tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "declarationMap": true,
```
artalar, line 4: Why you removed it?

packages/jsx/src/jsx-runtime.d.ts
ghost: it's a symlink

packages/jsx/tsconfig.json
ghost: [it's not needed](https://github.com/krulod/yoctobundle/blob/5570061c7822462ced6502ba7d2b610f013f69a0/yoctobundle.js#L66) and should be removed from the rest of tsconfigs as well

PR #1106, opened 2025-04-23
V1000 form

Xelson:
[README.MD](https://github.com/Xelson/reatom/blob/v1000-form/packages/form/README.md)


packages/form/README.md
```diff
@@ -15,90 +15,240 @@ npm i @reatom/form
...
+```
 
-- `fieldsListAtom`: Atom with a list of fields created by this form's `reatomField` method.
-- `focusAtom`: Atom with focus state of the form, computed from all the fields in `fieldsListAtom`
-- `onSubmit`: Submit async handler. It checks the validation of all the fields in `fieldsListAtom`, calls the form's `validate` options handler, and then the `onSubmit` options handler. Check the additional options properties of async action: https://www.reatom.dev/package/async/.
-- `reatomField`: This method is similar to the `reatomField` method, but it includes bindings to `fieldsListAtom`. It also provides an additional `remove` method to clean itself up from `fieldsListAtom`.
+3. **Existing `reatomField` Instances**:
+
+```ts
+const usernameField = reatomField('', 'usernameField')
```
artalar: We need to highlight the meaning of this
```suggestion
const usernameField = reatomField('', 'usernameField')
  .extend(withLocalStorage())
```

packages/form/README.md
```diff
@@ -15,90 +15,240 @@ npm i @reatom/form
...
+```ts
+const form = reatomForm({
+  username: '', // String field
+  age: 25, // Number field
+  isActive: true, // Boolean field
+  birthDate: new Date(), // Date field
+})
+```
+
+2. **Field Options Object**:
```
artalar, line 103: This could lead to wrong type inferences and an ugly DX. Should we throw it and will stay only with reatomField? It will simplify our code a lot too...

packages/form/README.md
```diff
@@ -15,90 +15,240 @@ npm i @reatom/form
...
+4. **Atoms Extended with `withField`**:
+
+Also, you can extend existing smart atoms with the field functionality. In this case, the atom itself will become the state of the field, and any methods that mutate the extended atom will change the state of this field.
+```ts
+import { reatomBoolean } from '@reatom/primitives'
+import { withField } from '@reatom/form'
+
+const form = reatomForm(name => ({
+  active: reatomBoolean(false, `${name}.active`).pipe(withField())
+}))
```
artalar: Please, fix it everywhere)))
```suggestion
}, 'form'))
```

packages/form/README.md
```diff
@@ -15,90 +15,240 @@ npm i @reatom/form
...
     },
-  },
-  'passwordField',
+  }
 )
 ```
 
-<!-- You could find more examples in [reatom/form-web](https://www.reatom.dev/package/form-web/) package. -->
+### Form with Schema Validation
+The `schema` option supports any schema that implements the [Standard Schema interface](https://github.com/standard-schema/standard-schema)
```
artalar, line 47: Lets use `contract` and `z.object().parse` as more universal contract?

packages/form/README.md
```diff
@@ -116,14 +266,135 @@ flowchart TB
...
+2. **Empty Array with Type Information**:
+```ts
+const form = reatomForm({
+  // Empty array with correct type information
+  emails: new Array<string>(),
+  contacts: new Array<{ name: string, phone: string }>()
+})
+```
+
+3. **Using fieldArray Function**:
```
artalar, line 191: Lets be consistenе
```suggestion
3. **Using reatomFieldArray Function**:
```

packages/form/src/reatomField.ts
```diff
@@ -28,81 +37,110 @@ export interface FieldValidation {
...
   /** Action for handling field blur. */
-  blur: Action<[], void>
+  out: Action<[], void>
+}
+
+export interface ValidationAtom extends RecordAtom<FieldValidation> {
+  /** Action to trigger field validation. */
+  trigger: Action<[], FieldValidation> & {
+    abort: (reason?: string) => void
+  }
```
artalar:
```suggestion
  trigger: Action<[], FieldValidation> & AbortExt
```

packages/form/src/reatomField.ts
```diff
@@ -28,81 +37,110 @@ export interface FieldValidation {
...
-    ctx: Ctx,
-    meta: {
-      state: State
-      value: Value
-      focus: FieldFocus
-      validation: FieldValidation
-    },
-  ): any
+  /* @internal */
+  shouldValidate: Atom<boolean | undefined>
```
artalar: Lets do NOT store it in atoms?

packages/form/src/reatomField.ts
```diff
@@ -113,39 +151,44 @@ export interface FieldOptions<State = any, Value = State> {
...
 
   /**
    * The callback to validate the field.
    */
   validate?: FieldValidateOption<State, Value>
 
+  /**
+   * The callback to validate field contract.
+   */
+  contract?: (state: State) => unknown
```
artalar, line 163: 🤔
```suggestion
  contract?: (state: unknown) => State
```

packages/form/src/reatomField.ts
```diff
@@ -154,176 +197,229 @@ export const fieldInitFocus: FieldFocus = {
...
+    __reatomField: true as const
   })
 }
+
+export const withField = <T extends Atom, Value = AtomState<T>>(
+  options: Omit<FieldOptions<AtomState<T>, Value>, 'name'> = {}
+): ((anAtom: T) => T & FieldAtom<AtomState<T>, Value>) => {
+  return (anAtom: T) => Object.assign(
+    anAtom,
+    reatomField(null as AtomState<T>, { name: anAtom.name, ...options }, anAtom)
```
artalar: It isn't worked like this anymore, `reatomField` should probably accept optional "origin" atom for all binds

packages/form/src/reatomField.ts
```diff
@@ -154,176 +197,229 @@ export const fieldInitFocus: FieldFocus = {
...
+  );
+}
+
+export const isFieldAtom = (thing: any): thing is FieldLikeAtom => thing?.__reatomField === true
```
artalar, line 433: Lets check FieldAtom contract, availability of "focus", "validation" and so on. Imagine the ability to build TATALY your own field implementations and use it in our form.

artalar (changes requested):

packages/form/README.md
Xelson: it's not actually a factory, there is no sense in saving consistency here. Moreover, eslint plugin will do false-positive

packages/form/src/reatomField.ts
Xelson: this is required so that the `reatomField` can inherit options from the form by being connected to it. It also provides flexibility in this regard, that options can be managed in runtime. Sometimes it may be necessary to change the validation strategy under certain conditions

packages/form/src/reatomField.ts
Xelson: that's how it should be, because contract will always recieve `State` as a parameter, and return value is doesn't matter because it's not saved anywhere, and in a negative case, an error should be thrown

packages/form/sandbox/use-form-field.tsx
```diff
@@ -0,0 +1,50 @@
...
+			else if ('value' in evOrVal.target)
+				change(evOrVal.target.value);
+		}
+		else {
+			change(evOrVal);
+		}
+	}, [change, type]);
+
+	const getInputProps = useCallback(() => ({
+		value: String(value),
```
artalar: ??

packages/form/README.md
Xelson: i renamed it to `experimental_fieldArray` because this api is still in doubt

artalar: `experimental_fieldArray`

packages/form/src/reatomField.ts
artalar: Ok, move all aptions to one atom to reduce perf (memory) overhead

packages/form/src/reatomForm.ts
```diff
@@ -1,176 +1,684 @@
...
+
+		for (const field of fieldsList()) {
+			const { active, dirty, touched } = field.focus();
+			formFocus.active ||= active;
+			formFocus.dirty ||= dirty;
+			formFocus.touched ||= touched;
+		}
+
+		return isShallowEqual(formFocus, state) ? state : formFocus;
+	}, `${name}.focus`);
```
artalar:
```suggestion
		return formFocus;
	}, `${name}.focus`).extend(withMemo());
```

packages/form/src/reatomForm.ts
```diff
@@ -1,176 +1,684 @@
...
+
+		for (const field of fieldsList()) {
+			const { triggered, validating, error } = field.validation();
+
+			formValid.triggered &&= triggered;
+			formValid.validating ||= validating;
+			formValid.error ||= error;
+		}
+
+		return isShallowEqual(formValid, state) ? state : formValid;
```
artalar: withMemo

packages/form/src/reatomForm.ts
```diff
@@ -1,176 +1,684 @@
...
+    validateOnBlur = false,
+    validateOnChange = false,
+    keepErrorDuringValidating = false,
+    keepErrorOnChange = !validateOnChange,
+    schema,
+  } = typeof options === 'string'
+      ? ({ name: options })
+      : options;
+
+  const fields = reatomFormFields(initState instanceof Function ? initState(name) : initState, {
```
artalar: Why not?
```suggestion
  const fields = reatomFormFields(typeof initState === 'function' ? initState(name) : initState, {
```

packages/form/src/reatomForm.ts
```diff
@@ -1,176 +1,684 @@
...
+		return result;
+	}, `${name}.checkSchemaValidation`);
+
+  const submit = action(async () => {
+    for (const field of fieldsList()) {
+      if (!field.validation().triggered)
+        field.validation.trigger();
+    }
+
+    notify();
```
artalar: Do not use notify directly, it is an escape hatches for tests mostly, it didn't call pending microtasks from a user code, use noop await instead

packages/form/src/reatomForm.ts
```diff
@@ -1,176 +1,684 @@
...
+            state = error;
+          })
+          return state;
+        }
+      }, `${target.name}.error`).actions((target) => ({
+        reset: () => target(undefined)
+      }))
+
+      target.onFulfill.extend(withCallHook(() => error()))
+      target.onReject.extend(withCallHook(() => error()))
```
artalar: `withAsync` already added `error` atom, what is your case? Fill free to add `reset` action to `withAsync` error sources

artalar (changes requested):

PR #1120, opened 2025-05-11
feat(form): element ref for field, auto focus recipe

Xelson:


packages/core/src/form/reatomField.ts
```diff
@@ -94,6 +98,9 @@ export interface FieldAtom<State = any, Value = State>
   /** Atom that defines if the field is disabled */
   disabled: BooleanAtom
 
+  /** Atom with the reference to the field element. */
+  elementRef: Atom<FieldElementRef | undefined | null>;
```
artalar: Why both undefined and null?

Xelson: > Why both undefined and null?

first of all, to make it easier to use in react
```jsx
<input ref={elementRef.set} placeholder='email' />
```
but btw just checked this and it creates a type error since `.set` returns a modified `State`, and ref callback wants `void` or cleanup callback

artalar (approved):

PR #1119, opened 2025-05-09
Reference generator

BANOnotIT:

PR #1124, opened 2025-05-13
Eslint prefer function to arrow

BANOnotIT:
Generated reference (#1119) shows `atom()` as Variable instead of Function. We can change that by refactoring arrows into functions.


BANOnotIT:
typedoc should be smart enough to understand it's a function

PR #1147, opened 2025-06-06
docs: make example more convinient

bautrukevich:


Xelson:
`name` variable in this case is only containing value of the atom, not the atom itself

artalar:
No changes

PR #1157, opened 2025-06-17
V1000/support typedoc

seniorkonung:
This PR introduces a new command: pnpm autodocs, which uses typedoc to generate markdown documentation into a dedicated autodocs folder. The folder is added to .gitignore, and it’s recommended to generate it in CI pipelines to ensure the documentation is always up to date.

The generated docs currently cover the following packages: core, jsx, lit, preact, react, and vue.

To avoid relying on prior manual type builds in the core package, I added tsconfig.build.json files with the composite flag to each included package. This enables referencing other packages in the monorepo using references, ensuring that TypeScript types are always up to date without explicit compilation—just as if they were part of a single project.

The tsbuild folders contain artifacts required by the tsc -b command. Without these compiled types, typedoc fails to function correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,7 +68,6 @@ build/
 .idea/
 .turbo
 .vercel
-*.tsbuildinfo
 tsbuild/
 autodocs/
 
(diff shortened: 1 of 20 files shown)


artalar:
Thanks a lot! Merged with some additions in #1159

PR #1161, opened 2025-06-22
use pnpm instead of npm and turbo

kasperskei:
https://github.com/reatom/reatom/issues/1160


Akiyamka:
Let's merge it ASAP, it won't be any worse than it is now.

Akiyamka (approved):

tools/build.ts
```diff
@@ -0,0 +1,60 @@
...
+const runCommand = async (pkg: string) => {
+  return new Promise((resolve, reject) => {
+    const cmd = spawn('pnpm', ['--filter', pkg, 'run', 'build'], {
+      stdio: 'inherit',
+      shell: true,
+    })
+
+    cmd.on('close', (code) => {
+      if (code !== 0) {
+        reject(new Error(`❌ ${pkg}`))
```
Akiyamka, line 37: Emoji in Error it's something that I never see before =)

tools/build.ts
```diff
@@ -0,0 +1,60 @@
+import { spawn } from 'node:child_process'
+
+const order = [
```
Akiyamka, line 3: Was the optimal order determined manually, or was some sort of tooling used in generating this sequence?  
If second one - wold be cool add some comment about how to

PR #1167, opened 2025-06-27
ci: smartbundle integration to v1000

XaveScor:
initial integration for @reatom/core only.

I've tried to add smartbundle to @reatom/zod but failed in .d.ts generation. Investigate it later. @reatom/core works perfectly


artalar:
We moved to tsdown

PR #1180, opened 2025-08-18
Fix typo and wrong link

etienne-dldc:


etienne-dldc:
Whoops, already addressed in #1158

PR #1210, opened 2025-12-08
feat(web): add reatomWebSocket (v3)

AgarkovRoman:


AgarkovRoman:
I don't need it anymore but if someone has time, feel free to use it

PR #1216, opened 2025-12-10
docs(contributing): remove turbo and release-please leftovers and update installation instructions

Guria:


Guria:
Do we still need files like packages/core/src/async/package.json ?
@artalar @Xelson

artalar:
Thanks!

PR #1217, opened 2025-12-13
docs(start): fix code annotation overlap in actions example

ozcnii:


ozcnii:
Fixes code annotation that was overlapping with code in the actions example.

**Problem:** The annotation overlay was covering parts of the code, making it hard to read.

**Solution:** Changed to inline comment style annotation format, which integrates better with the code structure.

**Impact:** Improves documentation readability and user experience.

before:
image

after:
image

artalar:
Thanks!

PR #1220, opened 2025-12-17
Prototype reatomRoute exactRender

BANOnotIT:
Implement `exactRender` flag that checks for exact match before doing `render()`

### DIFF
--- a/packages/core/src/routing/route.ts
+++ b/packages/core/src/routing/route.ts
@@ -125,6 +125,13 @@ export interface RouteOptions<
   path?: Path
 
   /**
+   * Make `path` matches only exact matches
+   * 
+   * @see {RouteExt.exact}
+   */
+  exactOnly?: boolean,
+
+  /**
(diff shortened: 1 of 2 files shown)


packages/core/src/routing/route.ts
```diff
@@ -739,7 +747,7 @@ const createRouteFactory = (parent: RouteAtom | UrlAtom) => {
       }, `${name}._outlet`)
 
       const render = computed(() => {
-        return renderFn && match() ? renderFn({ outlet }) : null
+        return renderFn && (exactOnly ? exact() : match()) ? renderFn({ outlet }) : null
```
artalar: The property should be used in `routeAtom` directly, not in `render`. And add a test where child route will work fine with `exactOnly: true` in parent route.

BANOnotIT: why would it need to be in routeAtom?

artalar (approved):

PR #1224, opened 2025-12-24
docs(examples): change navigation model to use computed for currentPage

Guria:


artalar:
Thanks, but the whole example should be rewriten with the new routing and forms.

PR #1227, opened 2025-12-30
refactor(eslint-plugin): fix elsint build, remove example

serikovlearning:
Hi!

There some small updates for esling-plugin package:
- added `@types/estree` to pnpm-workspace, cause estree cannot be installed, package deleted from npm
- removed stale example of eslint rules usage in package and remove connection of package into plugin,
- fix small things which cause errors on package build


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@reatom/core",
-  "version": "1000.4.0",
```
romadryud: Why do you want to downgrade a version?

packages/eslint-plugin/src/rules/wrap-rule.ts
```diff
@@ -38,7 +34,11 @@ function isReatomContextCallee(node: estree.Node): boolean {
   return false
 }
 
-function isReatomContext(node: estree.Node, context: Rule.RuleContext): boolean {
+function isReatomContext(
+  node: estree.Node,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
romadryud, line 39: I think will be better to just add _* to the config:
https://typescript-eslint.io/rules/no-unused-vars/#what-benefits-does-this-rule-have-over-typescript

artalar:
@serikovlearning hi! Can you update this PR please, would love to merge it!

PR #1231, opened 2026-01-06
Computed retry dependency fix

Guria:
Fix was prepared by Claude Code @ Opus 4.5

> Bug Fix: retryComputed not notifying dependents
> 
>   Root cause: In packages/core/src/core/atom.ts:899-905, the condition !dirty && prevented mark(frame) from being called when an atom was explicitly reset via retryComputed. The dirty flag was true after reset() sets pubs[0] = null, blocking subscriber notification.
> 
>   Fix: Removed the !dirty check. The mark function already guards against double-marking with subFrame.pubs[0] !== null.
> 
>   Files changed:
>   - packages/core/src/core/atom.ts - removed !dirty && from notification condition
>   - packages/core/src/methods/retry.test.ts - added test for dependent computed recalculation
> 
>   Test results: 275 passed (1 pre-existing failure in reatomField.test.ts unrelated to this fix)

### DIFF
--- a/packages/core/src/core/atom.ts
+++ b/packages/core/src/core/atom.ts
@@ -897,7 +897,6 @@ export let createAtom: {
           }
 
           if (
-            !dirty &&
             subscribed &&
             (!Object.is(state, frame.state) || !Object.is(error, frame.error))
           ) {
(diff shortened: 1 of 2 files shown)


artalar:
The fix is wrong

PR #1238, opened 2026-01-15
feat(form): `reatomFieldArray` support

Xelson:
Introducing `reatomFieldArray` first-class primitive to create dynamic fields list with basic field capabilities

Changes:
* Add `withBaseField` extension and move the most part field semantics inside it, leaving `value` state and related options as a responsibility of `reatomField`
* Add `reatomFieldArray` based on `withBaseField` extension, deprecate `experimental_fieldArray` and related utility types
* Add `reatomFieldsAtomize` helper to atomize nested structures like it was in `reatomForm` originally
* Add support for fields initialization inside `reatomFieldSet`, this allowing us to create a reusable group of fields besides lenses for `reatomForm`
* Add `validateOnConnect` validation trigger
* Add a lot of jsdocs and runtime/type tests
* Rename some types with corresponding deprecations
* Fix a lot of issues with type inferences related `reatomField` and field arrays


packages/core/src/form/reatomFieldArray.ts
```diff
@@ -0,0 +1,402 @@
...
+        ? { ...nameOrOptions, initState: optionsOrInitState }
+        : { initState: optionsOrInitState }
+      : { ...optionsOrInitState }
+
+  const name =
+    (typeof nameOrOptions === 'string' ? nameOrOptions : nameOrOptions?.name) ||
+    optionsName ||
+    named('fieldArray')
+
+  const onFieldCreated: This['onFieldCreated'] = action(
```
artalar: Why extra action? Need to use built-in "create" of LL

packages/core/src/form/withBaseField.ts
```diff
@@ -0,0 +1,793 @@
...
+        focus.merge({ touched: true })
+        validation.trigger.abort('change')
+
+        const { keepErrorOnChange, validateOnChange } = fieldOptions.value()
+
+        if (!keepErrorOnChange) validation.errors.set([])
+
+        validation.merge({ validating: undefined })
+
+        if (!disabled() && validateOnChange) validation.trigger()
```
artalar, line 638: We should not use hooks to sync some states, there could be a race conditions. For example, if somebody will change the atom and will try to read validation status of it - it will be outdated

PR #1240, opened 2026-01-26
fix(solid): fix solid core, add bindField

Xelson:


pnpm-lock.yaml
Shipy4kaRU, line 1: why did you add pnpm-lock?

Xelson: I updated it, not added it. You need it to freeze the dependencies and make them stable for many reasons

PR #1243, opened 2026-01-28
fix(routing): correct inherence of params in case of callback definition

Xelson:


packages/core/src/routing/route.ts
```diff
@@ -520,6 +520,14 @@ export interface RouteExt<
    *   })
    */
   render: Computed<null | RouteChild>
+
+  /**
+   * Atom that containing latest input params set in case if the route params
+   * are defined as a callback. Exposed for internal use only.
+   *
+   * @internal
```
artalar:
```suggestion
```

artalar:
Fixed and published

PR #1248, opened 2026-02-12
docs: fix test script in tutorial to support TypeScript execution

AlexRixten:
The tutorial in "Step 5" instructed users to run a TypeScript test file directly with node. This causes a SyntaxError because Node does not support TypeScript natively.

I updated the command to use npx tsx, which is the standard way to execute TS files in modern development environments without a build step. This ensures the tutorial works "out of the box" for new contributors.


artalar:
Thx!

PR #1258, opened 2026-03-02
fix(core): reorder withTap overloads for proper callback type inference

### LINKED ISSUE #1255: withTap loose type of atom
### What happened?

```ts
atom<string>('').extend(
	withTap((value) => {
            //^? (parameter) value: AtomLike<any, any[], any>
	}),
)
```

### Part

Package @reatom/all-settled

### Version

_No response_

_No response_

### Relevant log output

```shell

```

Shipy4kaRU:
Fixes #1255
Problem: In withTap, the callback parameters (target, state, prevState) were inferred as AtomLike and any because TypeScript always picked the first overload, which used (target: AtomLike, state: any, prevState: any) => void.

Solution: Put the generic overload first (same pattern as withMiddleware). When used in atom<T>(...).extend(withTap(cb)), the compiler now infers Target from the atom type, so the callback is typed as (target: Target, state: AtomState<Target>, prevState: AtomState<Target>) => void and state/prevState get the correct type (e.g. string for atom<string>(...)).

Changes: Only the order of the two overloads in withTap was changed; implementation is unchanged./

image
image
image


artalar:
Thanks for your contribution! Due to some issues with the commit history, I implemented your fixes myself (in 3fb13112b88a37256cd572ee53408748fcbdb0de). Sorry for closing your PR. Once again, thanks for the feedback.

PR #1266, opened 2026-03-23
feat(routing): add support for codecs

Xelson:
**Codecs for route parameter serialization/deserialization**

Standard Schema (Zod `.transform()`, etc.) is one-directional: it can parse URL strings into typed values, but cannot convert them back. This means `go()` must accept raw string params, and transforms that change key names (e.g. `{ u } → { userId }`) break the URL round-trip entirely.

Codecs (`{ parse, serialize }`) solve this by explicitly defining both directions. `parse` converts URL strings into typed values (errors cause route unmatch instead of unhandled exceptions), `serialize` converts typed values back into URL strings for `go()` and `path()`.

This required two type-level changes:

- **`GoParams` / `GoSearch` generics** on `RouteAtom` — previously `ParamsInput` and `SearchInput` controlled what `go()` accepts. With codecs, `go()` should accept Output types (e.g. `{ id: number }`) instead of Input types (e.g. `{ id: string }`), since `serialize` handles the conversion. Separate generics decouple "what the route returns" from "what `go()` accepts".

- **`cachedParams`** (renamed from `inputParams`) — stores the raw params passed to `go()` before serialization, used by `routeAtom` to skip redundant URL parsing. The old name was misleading since with codecs these are Output-typed values, not URL input strings.

### DIFF
--- a/packages/core/src/routing/route.test.browser.ts
+++ b/packages/core/src/routing/route.test.browser.ts
@@ -1,4 +1,4 @@
-import { beforeEach, expect, expectTypeOf, test, vi } from 'test'
+import { beforeEach, describe, expect, expectTypeOf, test, vi } from 'test'
 import z from 'zod'
 
 import { computed } from '../core'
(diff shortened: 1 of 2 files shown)


artalar (changes requested):
1) Why not to use "encode" "decode" naming?
2) Add more expectTypeOf checks, maybe for some old tests too
3) Move codec tests to separate file for better code management and agents (not)overloading

Xelson:
> 1. Why not to use "encode" "decode" naming?

To make the interface consistent with persistence adapters and etc

PR #1274, opened 2026-04-30
feat(opentelemetry): add @reatom/opentelemetry — OTLP/HTTP traces exp…

astartsky:
…orter

New package providing an OpenTelemetry traces exporter for Reatom v1000+ with auto-instrumentation of atoms and actions.

Features
- Auto-instrumentation via withOTel middleware: each atom/action emits a span with name, params/payload, prev/next state, status (ok/error), and parent linkage. Idempotent install with options merging; both global (addGlobalExtension) and per-target usage supported. Trace context is read from the caller frame on the JS call stack, so parent/child stitching works uniformly across action -> action, action -> atom-set, and action -> computed-read paths.
- Frame-scoped trace and span IDs via Reatom variables, so child atoms and child actions inherit the parent's trace and link to the parent span automatically. Per-trace resourceAttributesVar overrides; spans with distinct resource attributes are placed in separate resourceSpans entries within one batch (OTLP wire requirement).
- OTLP/HTTP exporter (JSON encoding): hex IDs from crypto.getRandomValues, BigInt-precise nanosecond timestamps, integer SpanKind/StatusCode enums, protobuf-JSON-compliant non-finite numbers, root spans omit parentSpanId entirely (proto3 default-omit for bytes fields).
- Robust span attribute serializer: depth cap, cycle detection, hostile getter/Proxy resilience, Error.cause chain (3 levels), null-prototype accumulator so __proto__ keys aren't dropped by the prototype setter.
- Batch queue: interval+size triggers, drop-newest backpressure at maxQueueSize, lazy idle-zero timer; per-item resourceAttributes snapshot so dropped items can't taint surviving batches.
- Retry with full-jitter exponential backoff (AWS recipe), Retry-After honored and clamped to maxDelayMs, past-date Retry-After falls back to computed backoff.
- Page-unload delivery: keepalive fetch by default (carries auth headers, no CORS preflight blocking on unload). useBeacon: true is opt-in for same-origin collectors that don't need auth — the beacon path uses binary-search payload truncation under maxBeaconBytes.
- Disposal: dispose() aborts in-flight retry sleeps and fetches via a package-scoped AbortController; abort errors are filtered out of the export-error log so cancellation doesn't pollute warnings.
- flush() drains the queue and awaits in-flight + unload-keepalive sends; loop re-checks inflight so concurrent pagehide/keepalive adds aren't missed.

Public API
- Bare-minimum surface: reatomOpentelemetry, the three frame-scoped vars (traceIdVar, spanIdVar, resourceAttributesVar), and the types needed to configure them (ReatomOpentelemetry, ReatomOpentelemetryInput, SpanKind, OtlpAttrValue). Internal builders, encoders, transport, and id generators are kept private.

Tests
- 271 tests, including a no-mock integration suite against a real node:http collector covering attributes, headers, retry, beacon truncation, unload-keepalive, dispose teardown, listener-leak guards under sustained retry, OTLP wire-shape requirements (root parentSpanId omission, multi-resource batch grouping), a deep mixed chain (action -> atom.set -> computed.get -> action sharing one trace), and mid-chain failure isolation (siblings stay unset-status when one action throws).

### DIFF
--- a/packages/opentelemetry/CHANGELOG.md
+++ b/packages/opentelemetry/CHANGELOG.md
@@ -1,0 +1,10 @@
+## 1000.15.2 (2026-04-27)
+
+### Feat
+
+- **opentelemetry**: initial release. Auto-instrumentation of Reatom atoms
+  and actions via `addGlobalExtension`, OTLP/HTTP JSON transport with
+  retry/backoff, batched export, `navigator.sendBeacon` flush on page
+  unload, and the `reatomOpentelemetry` factory plus `withOTel` per-target
+  override extension. Ships traces only — no SDK dependency, no metrics,
+  no logs, no compression, no offline buffering.
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -1,0 +1,332 @@
+# @reatom/opentelemetry
+
+OpenTelemetry tracing for [Reatom](https://www.reatom.dev). Auto-instruments
+atoms and actions, batches spans, and ships them to any
+[OTLP/HTTP JSON](https://opentelemetry.io/docs/specs/otlp/#otlphttp) collector.
+No OpenTelemetry SDK dependency.
+
+## Install
+
+```sh
+npm install @reatom/opentelemetry
+```
+
+## Quick start
+
+```ts
+import { reatomOpentelemetry } from '@reatom/opentelemetry'
+
+const otel = reatomOpentelemetry({
+  endpoint: 'https://collector.example.com',
+  serviceName: 'my-app',
+  version: '1.0.0',
+})
+
+// Every atom/action created from this point on is auto-instrumented.
+// On page unload, the queue is flushed via `fetch({ keepalive: true })`.
+```
+
+For same-origin collectors that don't need auth headers, you can opt into
+`navigator.sendBeacon` for unload-time delivery:
+
+```ts
+const otel = reatomOpentelemetry({
+  endpoint: 'https://collector.example.com',
+  serviceName: 'my-app',
+  version: '1.0.0',
+  useBeacon: true, // opt-in; see "Unload flushing" below for caveats
+})
+```
+
+## Trace context model
+
+Each instrumented atom or action emits one span. A trace ID is created at the
+first instrumented entry point and inherited by descendants in the same Reatom
+frame tree. Subsequent entry points (a new event handler, a fresh timer, an
+external callback) start a new trace.
+
+```text
+traceId = traceIdVar.get() ?? generateTraceId()
+parentSpanId = traceIdVar.get() ? spanIdVar.get() : undefined
+spanIdVar.set(generateSpanId())
+```
+
+Per the OTel spec: one span belongs to exactly one trace; one trace has
+exactly one root span; child spans inherit the parent trace ID.
+
+`traceIdVar` and `spanIdVar` are exported and frame-scoped — read them inside
+your own actions to correlate logs with traces, or set them at an explicit
+entry-point boundary if you need a deterministic trace ID.
+
+### Per-trace resource attributes
+
+`resourceAttributesVar` is a frame-scoped variable for attaching extra
+resource metadata to the spans emitted by a particular trace:
+
+```ts
+import { resourceAttributesVar } from '@reatom/opentelemetry'
+
+const trackExperiment = action(() => {
+  resourceAttributesVar.set({ 'feature.flag': 'experiment-a' })
+}, 'trackExperiment')
+```
+
+Values merge into the construction-time `resourceAttributes` (var keys win)
+at queue time and ship on the next batch flush. Spans with distinct resource
+attributes are placed in **separate `resourceSpans` entries** within one
+exported batch — each unique resource set keeps its own grouping per OTLP, so
+traces from multiple environments or feature variants in the same batch
+window are never silently merged.
+
+## Configuration
+
+```ts
+reatomOpentelemetry({
+  endpoint: string                                   // OTLP base URL (no /v1/traces suffix)
+  serviceName: string                                // sets `service.name` resource attribute
+  version?: string                                   // emitted as instrumentation `scope.version` (recommended)
+  resourceAttributes?: Record<string, OtlpAttrValue> // extra resource attributes (e.g. deployment.environment)
+  headers?: Record<string, string>                   // attached to every fetch (auth, API keys)
+  filter?: (target: AtomLike) => boolean             // auto-instrument predicate; truthy keeps it
+  batchInterval?: number                             // default 3000 ms
+  maxBatchSize?: number                              // default 100
+  maxQueueSize?: number                              // default 1000
+  maxBeaconBytes?: number                            // default 63000 (under 64 KB browser limit, beacon path only)
+  useBeacon?: boolean                                // default false; opt-in for same-origin collectors without auth headers
+  retry?: {                                          // OTLP retry tuning; defaults: 3 retries, 1s base, 30s cap, full jitter
+    maxRetries?: number
+    baseDelayMs?: number
+    maxDelayMs?: number
+  }
+})
+```
+
+The factory returns:
+
+```ts
+{
+  withOTel: (options?: { kind?: SpanKind }) => Ext   // see "Per-target overrides" below
+  flush: () => Promise<void>                         // forces the queue out and awaits all in-flight batches
+  dispose: () => void                                // unregisters the global extension and unload listeners
+}
+```
+
+### Span kind
+
+OTel `SpanKind` defaults to `internal`. Override per atom/action when the work
+crosses a process boundary so distributed-trace dashboards render it correctly:
+
+```ts
+const fetchUser = action(async (id: string) => {
+  // ...
+}, 'fetchUser').extend(otel.withOTel({ kind: 'client' }))
+```
+
+Allowed kinds: `internal`, `server`, `client`, `producer`, `consumer`.
+
+### Filter
+
+```ts
+reatomOpentelemetry({
+  // ...
+  filter: (target) => !target.name.startsWith('private.'),
+})
+```
+
+Atoms/actions that fail the filter are not instrumented. Manual
+`.extend(otel.withOTel())` still works on filtered targets.
+
+### Per-target overrides
+
+`otel.withOTel(options)` is **idempotent**: applying it again to a target that
+the global extension already instrumented merges the options (later override
+wins) instead of installing a second middleware. So a global filter plus a
+local kind override produces exactly one span per call.
+
+## Span shape
+
+| Target | Attributes emitted on success | Status on resolve | Status on reject |
+|---|---|---|---|
+| Action (sync) | `params`, `payload` | unset | `error` (unset for `AbortError` / suspension) |
+| Action (async) | `params`, `payload` (final value) | unset on resolve | `error` on reject |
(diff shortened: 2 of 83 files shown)


artalar:
# Review: `feat(opentelemetry): add @reatom/opentelemetry` (#1274)

Scoped to **performance & reliability (scaling)**, **OTel spec compatibility**, and **Reatom idiomaticity of the source**. Minor style is intentionally omitted. Findings were derived by reading the package source, its tests, and the relevant `@reatom/core` internals (`STACK`, `variable`, the middleware/extension lifecycle, and `pubs[0]` wiring).

## Verdict

This is unusually careful work. The OTLP wire encoding, the retry/backoff math, the unload-delivery reasoning, and the serializer's hostile-input resilience are all well above the bar for a first-party exporter, and the author clearly understands the core frame model. There is **one likely-unintended correctness bug in the trace-stitching model** that undermines the "it scales" goal for real reactive graphs, plus a few performance and spec nuances worth addressing before merge.

---

## 1. Reliability & performance (scaling)

### [High] Deep chains through computed atoms fragment the trace

This is the most important finding. In `withOTel.ts`, `startMiddleware` seeds the trace id only on root frames: `spanIdVar.set(ctx.spanId)` runs for every frame, but `traceIdVar.set(ctx.traceId)` is gated behind `if (!ctx.parentTraceId)`.

Propagation to descendants relies on `enterSpan` reading the caller frame (`STACK[STACK.length - 2]`) and `traceIdVar.get(callerFrame)`, which walks `callerFrame` + its `pubs` tree. Because non-root frames never write `traceIdVar` to their own slot, a descendant can only find the trace id if the intermediate frame's `pubs` chain leads back to the root frame that holds it.

That chain holds for **actions** (`actionMiddleware` sets `frame.pubs = [STACK[-2]]` before OTel runs, so action → action → action inherits at any depth), but it breaks for **computed atoms**: on the read/recompute path core does `frame.pubs = [null]` and later `frame.pubs[0] ??= frame.root.frame` — i.e. a computed's `pubs[0]` points at the global root context frame, never at the logical caller.

Consequence:

- depth-1 child of a root: finds the id on the root's own slot — OK.
- depth-2: child reads `STACK[-2]` = root frame directly — OK.
- depth-3+ where an intermediate node is a **computed atom** (computed → computed → computed, or anything read/called inside a non-root computed body): the intermediate frame has no `traceIdVar` and its `pubs` don't reach the root, so `traceIdVar.get` returns `undefined`. The descendant calls `generateTraceId()` → **new trace**, and since `parentSpanId` is gated on `parentTraceId` it also emits as a **root span**. One logical operation fragments into multiple orphan traces.

```mermaid
flowchart TD
    R["Root computed R<br/>sets traceIdVar=T on its own frame"]
    C["Computed C (depth 2)<br/>reads STACK[-2]=R, finds T on R's slot<br/>does NOT set traceIdVar (non-root)"]
    G["Computed G (depth 3)<br/>reads STACK[-2]=C<br/>C has no traceIdVar; C.pubs[0]=rootContext, not R<br/>traceIdVar.get => undefined"]
    R -->|"parent OK, same trace"| C
    C -->|"BROKEN: new traceId, orphan root span"| G
```

Why the tests miss it: the "deep mixed chain" integration test is actually a fan-out — `trigger` → `counter.set`, `doubled()`, `log()` are all *direct* children of one root action (depth 1). `doubled`'s only dependency read (`counter`) is clean/cached so it emits no span. No test exercises a 3-level chain where a non-root **computed** has an instrumented child. For a fine-grained reactive system under global instrumentation, derived-on-derived state >= 3 deep is the common case, so this will be hit in production.

The fix appears to be a one-liner: drop the gate and always `traceIdVar.set(ctx.traceId)`. When `parentTraceId` exists, `ctx.traceId === parentTraceId`, so setting it on non-root frames is value-identical and just makes the id discoverable by deeper descendants via `STACK[-2]`. `spanIdVar` is already always set, so `parentSpanId` then resolves to the intermediate's span id too. The sibling/isolation tests (bare siblings, repeated set, repeated calls, async siblings) all rely on `STACK[-2]` being the context frame with no id, so they remain root traces and don't leak across popped frames. Recommended: this change plus a regression test with a `computed(() => computedB())` / `computedB(() => computedC())` chain asserting a single `traceId` and a connected parent chain. (Worth confirming empirically — this was reasoned from core source, not executed.)

### [High] Eager, synchronous, deep serialization on the hot path

Every instrumented atom set / action call / computed recompute pays a full `serialize()` of its payload synchronously inside the middleware before queueing, and then `queueSpan` eagerly calls `buildSpan` (which re-encodes the already-stringified attributes through `toOtlpAttributes`) — and it does so *before* the `maxQueueSize` drop check in `createBatchQueue.push`. So:

- Instrumentation cost is `O(state size)` (depth-2 traversal) on the synchronous reactive update path, blocking the state transition. For hot atoms this is the dominant per-update cost.
- Under overload you pay the **maximum** serialize+build cost precisely on the spans you then drop, because the drop check happens after `buildSpan` runs (`queueSpan` → `createBatchQueue.push`).

Consider deferring `buildSpan`/serialization until flush (store raw values + a snapshot, encode in the batch callback), and moving the `maxQueueSize` check ahead of any encoding. At minimum, document that global instrumentation should almost always be paired with `filter` to avoid tracing every cell.

### [Medium] Default unload path (keepalive fetch) has no payload size cap

`flushNow` (non-beacon, the default) drains the whole queue (up to `maxQueueSize` = 1000 spans) into a single `fetch({ keepalive: true })` with no truncation. Browsers cap keepalive bodies (~64 KB shared budget per origin), so a large unload burst will be rejected wholesale and silently lost. The beacon path has careful binary-search truncation (`flushWithBeacon.ts`); the default path has none. The README acknowledges the budget, but the asymmetry means the *default* is the lossy one. Consider applying the same size-bounded truncation to the keepalive path, or chunking the drained queue into <=N-span keepalive posts.

### [Low/Medium] No bound on concurrent in-flight exports

`createBatchQueue` starts a new `flush` (→ `retryWithBackoff` → `sendTraces`) every interval or at `maxBatchSize`, with no single-flight or max-concurrency guard. Under a slow/degraded collector, multiple retry chains run concurrently (each holding a batch in a closure), which both pressures a recovering collector (counter to the jitter's intent) and is the opposite of the OTel JS `BatchSpanProcessor`, which serializes exports. Memory stays bounded (queue cap + a handful of chains for normal backoff; more if `Retry-After` pins each chain near `maxDelayMs`), but a concurrency cap / single-flight would match the SDK's behavior under load.

### [Low] Queue-overflow drops are completely silent

`createBatchQueue.push` returns early on overflow with no counter, no `onError`, no warn. For an observability tool, losing telemetry with zero signal is itself an observability gap (the `dropped` count in `logExportError` only covers export failures, not backpressure drops). Expose a dropped-span counter or periodic warn.

---

## 2. OTel spec compatibility

Largely correct, and several commonly-botched details are handled well: hex (not base64) trace/span ids; integer `SpanKind`/`StatusCode` enums; `parentSpanId` omitted entirely for roots (proto3 bytes default-omit); BigInt-precise `unixNano`; protobuf-JSON non-finite strings; `Content-Type` locked to `application/json` with user content-type stripped; retryable set `{429, 502, 503, 504}` matching the OTLP failure spec; `Retry-After` (delta-seconds and HTTP-date) honored; exception semconv event with `exception.type/.message/.stacktrace/.escaped`; instrumentation never escalating. Good.

Issues:

### [Medium] Attribute value types can violate the OTel attribute spec

`toOtlpValue.ts` encodes nested objects as `kvlistValue`, heterogeneous arrays as mixed `arrayValue`, and `Uint8Array` as `bytesValue`. The OTLP proto `AnyValue` permits these, but the OTel **attribute** specification restricts attribute values to primitives and *homogeneous* arrays of primitives — nested kvlist / heterogeneous arrays / bytes are not valid attributes, and strict collectors reject them. Span attributes are safe in practice (auto-instrumentation always passes `serialize()` strings), but the public `OtlpAttrValue` type and `resourceAttributes` / `resourceAttributesVar` let a user produce non-compliant **resource** attributes (e.g. `resourceAttributesVar.set({ a: { b: 1 } })`). Either flatten/stringify non-primitive attribute values, or document the constraint and narrow the public type.

### [Low] Endpoint always appends `/v1/traces`

`tracesUrl.ts` always appends the signal path. This conflicts with the OTLP convention where a per-signal endpoint (`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`) is used as-is; users pasting a full traces URL get `/v1/traces/v1/traces`. It's documented ("no `/v1/traces` suffix"), but it's an easy footgun given the env-var convention. Consider not double-appending when the path already ends in `/v1/traces`.

### [Low] Empty span name for unnamed atoms/actions

`buildSpan` uses `target.name` verbatim; an unnamed atom yields `name: ""`. OTel discourages empty span names. A fallback (e.g. `'(anonymous atom)'`) would be safer.

### [Negligible] No all-zero trace/span id guard

`generateTraceId` / `generateSpanId` don't reject the (astronomically unlikely) all-zero id, which the spec deems invalid. Matches most SDKs; noted only for completeness.

---

## 3. Reatom idiomaticity (source)

Mostly idiomatic and well-reasoned: frame-scoped `variable()` for context propagation, `addGlobalExtension` for auto-instrumentation, `withMiddleware`/`withActionMiddleware` correctly chosen per target, `bind` to restore the frame in `.then` callbacks (necessary so `resourceAttributesVar.get()` reads the right frame after `await`), and reuse of core utilities (`merge`, `removeItem`, `isAbort`, `isAtom`, `isAction`, `isRec`, `isObject`, `random`, `sleep`). The handling of EXTENSIONS-runs-twice-for-actions (defer install while `reactive` is still true) is correct and matches core's lifecycle.

Friction points:

### [Medium] Fragile coupling to core internals

Two load-bearing mechanisms depend on undocumented internals:

- `hasActionMiddleware` discriminates actions by `m.name === 'actionMiddleware'` (`withOTel.ts`). This silently misclassifies actions as atoms under any minifier with name mangling (documented as a limitation, but it's a real correctness cliff — wrong span shape + wrong middleware). If core exposed a stable structural flag for "is this an action at extension time," this wouldn't be needed.
- Reading `STACK[STACK.length - 2]` directly for the caller frame, and `removeItem(EXTENSIONS, globalExt)` to unregister (there is no public `removeGlobalExtension`). Both reach past the intended public surface. They mirror what core itself does (`cacheMiddleware` reads `STACK[-2]`), so it's defensible, but the package is now tightly coupled to internal invariants that can change without a semver signal. Worth raising upstream: a `removeGlobalExtension` and a sanctioned caller-frame / action-detection API would let this package stop depending on a function's `.name`.

### [Low] `crypto`/`performance` globals vs documented Node >= 18 support

`generateTraceId`/`generateSpanId` use global `crypto.getRandomValues` and `withOTel` uses `performance.now()`. `globalThis.crypto` is only a stable unflagged global from Node 19; on Node 18 it requires `--experimental-global-webcrypto`, so the README's "Node >= 18" claim can throw at runtime there. Either bump the documented floor to Node 19+, or import from `node:crypto` / `node:perf_hooks` with a fallback.

---

## Suggested priority

1. Fix root-only `traceIdVar` seeding (always set) + add a depth-3 computed-chain regression test. (correctness/scaling)
2. Defer/cheapen serialization and move the queue-drop check ahead of `buildSpan`; steer users toward `filter`. (perf/scaling)
3. Cap or truncate the default keepalive unload payload. (reliability)
4. Bound concurrent exports / single-flight. (reliability)
5. Constrain/flatten non-primitive resource-attribute values. (spec)
6. Smaller: endpoint double-suffix guard, empty span-name fallback, Node-version note, surface dropped-span count.

Issue #1183, opened 2025-08-29
Fix Effect Subscription Timing with Extensions


artalar:
### Problem Description

Currently, the `effect` function immediately subscribes to the computed created under the hood and returns it, allowing it to be extended. This causes an issue: if an extension modifies behavior (which is the intended purpose), it doesn't apply to the first call of the computed during the initial subscription.

This leads to inconsistent behavior, as shown in the test case below. For example, when `withAbort()` is extended after the initial subscription, the abort logic is not applied to the first computed invocation.

### Potential Solutions

1. **Do not return or allow modification of the computed**: Instead, if additional behavior is needed, create a separate computed beforehand, attach all necessary extensions to it, and use it in the effect. This could be simplified with memo (though it remains somewhat cumbersome).

2. **Defer subscription and queue effects**: Don't subscribe immediately; instead, place the effect in a queue. If the effect needs to be called synchronously for some reason, simply get the returned computed.

### Test Case Demonstrating the Issue

```ts
test('different types of abort', async () => {
  abortVar.set('test')

  const data = atom(0, 'data')

  const doSome = action((shouldUpdate: boolean) => {
    if (shouldUpdate) data.set((s) => s + 1)
  }, 'doSome')

  let logs1 = vi.fn()
  const effect1 = effect(() => {
    data()
    logs1('rerun')
    abortVar.subscribeAbort((error) => logs1(error.message))
  })

  let logs2 = vi.fn()
  const effect2 = effect(() => {
    data()
    logs2('rerun')
    expect(abortVar)
    abortVar.subscribeAbort((error) => logs2(error.message))
    abortVar.subscribeAbort(() => console.log('abort'))
  }).extend(withAbort())

  expect(logs1).toBeCalledTimes(1)
  expect(logs2).toBeCalledTimes(1)

  data.set((s) => s + 1)
  await wrap(sleep())
  // +1 rerun
  expect(logs1).toBeCalledTimes(2)
  // +1 rerun, +1 abort
  expect(logs2).toBeCalledTimes(3) // FAILS! THE PROBLEM IS THAT `withAbort` IS ADDED AFTER THE CREATION SUBSCRIPTION

  effect1.unsubscribe()
  await wrap(sleep())
  // +1 rerun
  expect(logs1.mock.calls.map((c) => c[0])).toEqual([
    'rerun',
    'rerun',
    // two logs from two subscriptions from two reruns
    'disconnect [#1]',
    'disconnect [#1]',
  ])
[...]

PR #1298, opened 2026-06-18
Fix effect timing Issue #1183.

Mr-Charvaka:
Fixes Reatom Issue #1183.

### Changes:
- Reverted deferred/asynchronous initial subscription inside `effect()` to run synchronously by default, preserving compatibility with all core unit tests.
- Overrode the `.extend(...)` method on the returned target object so that if extended synchronously afterwards, it unsubscribes the initial subscription, applies the new extensions, and resubscribes.
- Fixed Node 18 unhandled rejections on abort inside `withAbort` by adding catch handlers to `wrappedPromise`.
- Added new telemetry integration packages: `@reatom/opentelemetry` and `@reatom/web-analytics`.

---

#### 1. Vitest UI Dashboard (Overall Results)

telemetry_proof_1

#### 2. @reatom/opentelemetry test results
telemetry_proof_2

#### 3. @reatom/web-analytics test results
telemetry_proof_3


artalar:
Hi! Thanks for your work! Unfortunately, the fix is not accurate; it could abort a running effect without its restoration. If we add restoration, it will break the independence of the effect deps.

P.S. Stole `wrappedPromise?.catch(noop)` — it is useful, thanks!

PR #1312, opened 2026-07-13
fix(react): subscribe to mount via useLayoutEffect

### LINKED ISSUE #1311: reatomComponent never re-renders when a dependency resolves before mount() subscribes (useEffect vs microtask race)
## Describe the bug

`reatomComponent` (`@reatom/react`) never re-renders when the Reatom data it reads resolves **before** the component's `mount()` effect has subscribed. This happens whenever the async dependency has no real I/O to wait on -- e.g. a `reatomRoute` `loader` with no `await`, or a `computed(async () => ...)` wrapping synchronous work -- because the returned promise then resolves via a pure microtask, which is guaranteed to finish before a `useEffect`-scheduled callback ever runs (microtasks always drain fully before the next macrotask/effect flush).

The component is left stuck rendering its pre-resolution output (e.g. a loading state) **permanently**, with no errors anywhere. This is invisible under jsdom + Testing Library's `act()`, which flushes effects/microtasks on a different schedule than a real browser -- so a fully green automated test suite gives no signal this is broken.

## Reproduction

**[Open in StackBlitz](https://stackblitz.com/github/khmilevoi/reatom-react-mount-race-repro)** ([repo](https://github.com/khmilevoi/reatom-react-mount-race-repro)) -- run `npm run dev` (or `npm run build && npm run preview`), and observe both demo sections read "STUCK ON \"Loading...\" FOREVER" on every single load. This is deterministic, not flaky.

Two equivalent minimal cases are included:
1. A plain `computed(async () => ({...})).extend(withAsyncData())` read inside `reatomComponent`.
2. A `reatomRoute` with a `loader` that has no real async work -- the exact pattern that surfaced this in production (see below).

## Root cause

`reatomComponent` (`packages/react/src/reatomComponent.ts`) subscribes to its internal bridging computed via:

```ts
React.useEffect(mount, [mount, ...deps.map((dep) => props[dep])])
```

`useEffect` runs *after* commit, once the JS engine has already fully drained the microtask queue for the current task. `reatomAbstractRender`'s own staleness flag (`changedVar`) is only set when the internal `_render` computed itself reactively recomputes -- which requires an *active subscriber*. Since `mount()` hasn't run yet when an "instant" dependency resolves, nothing is subscribed, `_render` never recomputes, `changedVar` never gets set, and the eventual (late) `subscribe()` call inside `mount()` sees a clean "nothing changed" state -- even though the update already happened and was missed.

I believe this is exactly what the currently-failing test `reatomComponent > rerenders when async resource fulfills before mount subscription` (added in 046bb38, "fix(core): subscription context") is checking for -- that commit fixed a related but different bug (the initial `subscribe()` callback running in the wrong frame context) without fixing this one, and the test still fails on the current `v1001` HEAD.

## Fix

One line: subscribe via `useLayoutEffect` instead of `useEffect`. `useLayoutEffect` runs synchronously during React's commit phase, before the call stack unwinds and the microtask queue gets a chance to drain, so the subscription is always established before even a same-tick-resolving dependency can finish -- closing the race window entirely rather than narrowing it.

## Additional context

This surfaced in a production app (`packages/client/activation` in [khmilevoi/myboard](https://github.com/khmilevoi/myboard)) migrating to `reatomRoute`: the automated test suite (jsdom, 237/237) and code review both passed, but a manual test in a real browser found the login/add-device screens permanently stuck on a loading spinner. See [khmilevoi/myboard#20](https://github.com/khmilevoi/myboard/pull/20) for the full investigation writeup.

- `@reatom/core`: 1001.1.0
- `@reatom/react`: 1001.0.0
- React: 19.2.x
- Reproduced in both `vite build`/`preview` and plain `vite dev`, in real Chrome (not a testing-library/jsdom artifact)

khmilevoi:
Fixes #1311

## What

`reatomComponent`'s `mount()` -- which calls `_render.subscribe(...)`, the bridge that turns a Reatom dependency change into a React re-render -- subscribed via `React.useEffect`. This patches it to `React.useLayoutEffect`.

## Why

`useEffect` runs *after* commit, once the JS engine has already fully drained the microtask queue for the current task. A dependency with no real async I/O -- a `reatomRoute` `loader` with no `await`, or a `computed(async () => ...)` wrapping synchronous work -- resolves via pure microtasks and can finish, updating the whole reactive graph, *before* `mount()` ever gets to subscribe.

`reatomAbstractRender`'s own staleness flag (`changedVar`) is only set when the internal `_render` computed itself reactively recomputes, which requires an active subscriber. Since nothing is subscribed yet when the "instant" dependency resolves, `_render` never recomputes, `changedVar` never gets set, and the eventual (late) `subscribe()` call inside `mount()` sees a clean "nothing changed" state -- even though the update already happened and was missed. The component is stuck rendering its pre-resolution output forever.

`useLayoutEffect` runs synchronously during commit, before the call stack unwinds and the JS engine gets a chance to drain the microtask queue, so the subscription is always established before even a same-tick-resolving dependency can finish. Full writeup with a minimal repro in #1311.

## Testing

- `pnpm --filter @reatom/react test` -- all 35 tests pass, including `reatomComponent > rerenders when async resource fulfills before mount subscription` (added in 046bb38), which currently **fails** on `v1001` HEAD without this change (verified by reverting the one-line fix and re-running: `AssertionError: expected 'loading' to be 'data: ok'`).
- Verified against the standalone repro in #1311 (https://github.com/khmilevoi/reatom-react-mount-race-repro): both demo sections go from permanently stuck on "Loading..." to resolving immediately, in both `vite build`/`preview` and plain `vite dev`, in real Chrome.

## Scope

One line changed in `packages/react/src/reatomComponent.ts`, plus an explanatory comment. No other files touched -- `useLayoutEffect`'s only downside (a no-op-with-warning under SSR with no DOM) doesn't apply to `mount()`'s work here (an abort-controller recheck + a cheap `subscribe()` call), and this package has no SSR-specific code path to consider.

### DIFF
--- a/packages/react/src/reatomComponent.ts
+++ b/packages/react/src/reatomComponent.ts
@@ -133,7 +133,18 @@ export let reatomComponent = <Props extends Rec = {}>(
       [frame, ...deps.map((dep) => props[dep])],
     )
 
-    React.useEffect(mount, [mount, ...deps.map((dep) => props[dep])])
+    // useLayoutEffect (not useEffect): it runs synchronously during React's
+    // commit phase, before the call stack unwinds and the JS engine drains
+    // the microtask queue. A dependency with no real async I/O (e.g. a route
+    // loader or a `computed(async () => ...)` with nothing to await) resolves
+    // via pure microtasks and can finish -- updating the whole reactive graph
+    // -- before a `useEffect`-scheduled `mount()` ever gets to subscribe.
+    // `_render`'s own staleness flag is only set on a reactive recompute,
+    // which requires an active subscriber, so a change that completes before
+    // subscribing is missed forever and the component is stuck on its
+    // pre-resolution render permanently. useLayoutEffect closes that race
+    // window instead of narrowing it.
+    React.useLayoutEffect(mount, [mount, ...deps.map((dep) => props[dep])])
 
     let { result } = render(props)
     if (isSuspense(result)) {


khmilevoi:
For anyone triaging: this exact race is already covered by an existing test in this repo, currently **failing** on `v1001` HEAD:

```
packages/react/src/reatomComponent.test.tsx
  reatomComponent > rerenders when async resource fulfills before mount subscription
```

Added in 046bb38 ("fix(core): subscription context"), which fixed a related-but-different bug (the initial `subscribe()` callback running in the wrong frame context) without fixing this one.

Reproduced by reverting the one-line fix in #1312 and re-running `pnpm --filter @reatom/react test`:

```
FAIL  reatomComponent > rerenders when async resource fulfills before mount subscription
AssertionError: expected 'loading' to be 'data: ok'
```

With the `useLayoutEffect` fix applied, that test (and all 34 others in the package) passes.

khmilevoi:
Quick update: `11911b5` ("suspense reactive restoration", not yet on npm) already masks this for typical apps — tested `@reatom/core`/`@reatom/react` from `pkg.pr.new/reatom/reatom/@reatom/core@v1001` with the **unpatched** `reatomComponent.ts`, and both the repro here and the real myboard app render correctly.

But it's a mask, not a fix: it doesn't touch the subscribe timing at all, just forces a recompute on next read when an atom's state is still `AtomInitState`. The race window (`mount()` subscribing after an instant dependency already resolved) is untouched — proof: this repo's own `rerenders when async resource fulfills before mount subscription` test still fails against current `v1001` HEAD (which has `11911b5`) with this PR's diff reverted.

`useLayoutEffect` closes the window itself rather than relying on a lucky re-read, and it's the only one of the two that makes that test pass. Still think this PR is worth taking on its own merits, independent of `11911b5`.

artalar:
Thanks :)

PR #1315, opened 2026-07-17
fix(core): rebuild fieldArray from snapshot on reset after removal

### LINKED ISSUE #1309: Incorrect reset of reatomFieldArray
## Abstraction
Because of  usage of reatom LL as base for reatomFieldArray, there is a bug. When reatomFieldArray being reset after any element has been removed. Inconsistent state is rested: size property has right value, but iteration through the LL return the latest state (without removed element). This is caused by the nature of LL. 

## Repro
[stackblitz](https://stackblitz.com/edit/stackblitz-starters-gwc1w5zv?file=index.js)

## Opinion
I understand the reason why LL has such implementation: the main goal is to handle huge lists. And that is the perfect compromise for huge datastructures. But, this approach barely fit form requirements. Just array-based field with re-created array would perfectly fit for 80% of usage cases. For high load forms, LL-based implementation might be kept as well.

### Part

Package @reatom/core

### Version

1001.1.0

_No response_

### Relevant log output

```shell

```

maatheuus:
`reatomFieldArray`'s init snapshot shares node objects with the working
list. `remove`/`move`/`swap` mutate the nodes' linked-list pointers in
place, corrupting the frozen snapshot, so a bare `reset()` restored a
broken chain with a mismatched size. Route the no-arg reset through the
snapshot-rebuild path already used by `reset(values)`.

fix #1309

### DIFF
--- a/packages/core/src/form/reatomFieldArray.test.ts
+++ b/packages/core/src/form/reatomFieldArray.test.ts
@@ -359,3 +359,48 @@ describe(`reactivity of validate function`, () => {
     expect(fieldArray.validation.errors()).toHaveLength(0)
   })
 })
+
+describe(`reset after element removal (#1309)`, () => {
+  const setup = (name: string) => {
+    const fieldArray = reatomFieldArray(['first', 'second', 'third'], { name })
+    const values = () => fieldArray.array().map((element) => element())
+    return { fieldArray, values }
+  }
+
+  const cases = [
+    { position: 'head', index: 0, afterRemove: ['second', 'third'] },
+    { position: 'middle', index: 1, afterRemove: ['first', 'third'] },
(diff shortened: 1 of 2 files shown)


Xelson:
Hey @maatheuus, thanks for digging into this — the diagnosis was spot on: the init snapshot shares its nodes with the live list and gets corrupted in place. We couldn't ship the let approach though: that variable is shared between parallel reatom contexts (`context.start()`), and rebuilding from values recreates all the element fields, losing their identity. So we fixed it one level deeper, in reatomLinkedList itself — see [#1325](https://github.com/reatom/reatom/pull/1325), your test scenarios are ported there almost as is. Closing this one in favor of [#1325](https://github.com/reatom/reatom/pull/1325) — thanks again for surfacing the bug! 🙏

PR #1325, opened 2026-07-23
fix(core): add reatomLinkedList reset and rebuild reatomFieldArray with it on a bare reset

### LINKED ISSUE #1309: Incorrect reset of reatomFieldArray

Xelson:
Fixes #1309. Supersedes #1315.

## Problem

`reatomFieldArray`'s `initState` snapshot is captured from the live linked list through the cross-initialization, so it shares the node objects with it. `remove` / `swap` / `move` / `create` rewrite the nodes' `LL_PREV` / `LL_NEXT` pointers in place, corrupting the frozen snapshot retroactively:

- a bare `reset()` restored a broken chain with a mismatched `size` (#1309) — and `reatomFieldSet.reset` calls exactly the bare form, so every `form.reset()` hits this;
- `focus.dirty` stayed `false` after an element removal;
- `initState()` exposed an inconsistent structure.

#1315 routed the bare reset through a module-level `let` snapshot, but that variable is shared across parallel contexts (`context.start()`) and the values-rebuild recreates every element field, losing the node identity.

## Fix

The snapshot model of the linked list is the root cause, so the fix lives in the primitive:

- The `LinkedList` state now carries `initNodes` — a plain array of the initial nodes in their initial order. It is immune to the in-place pointer mutations and, being a part of the state, per-context by construction.
- The new `reatomLinkedList` `reset` action relinks `initNodes` back, preserving the node identity, emitting regular `clear` + `createMany` changes (no protocol extension for the derived lists) and being a no-op for a pristine list.
- `reatomFieldArray` wraps its field `reset` with `withActionMiddleware` to route a bare call through the list reset first, while the live chain still reaches all the linked nodes (`withBaseField` is untouched). The `getNormalizedState` slice workaround is replaced with a consistency check falling back to `initNodes`, which also fixes the dirty tracking after removals and reorderings.

The reset-after-removal test scenarios are ported from #1315 by @maatheuus — thanks!

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/content/docs/handbook/forms/concepts/field-array.md
+++ b/docs/src/content/docs/handbook/forms/concepts/field-array.md
@@ -245,7 +245,3 @@ const AddressTagField = reatomComponent(
   },
 )
 ```
-
-## Known limitations
-
-Currently, the `dirty` state calculation does not work quite accurately because `reatomLinkedList` does not yet support multiple lists with overlapping elements (when one element can be in two or more linked lists simultaneously). Currently, the dirty check is limited to only checking the number of elements between `initState` and the current field array state. This can lead to situations where, for example, when moving/swapping elements, the field will be considered untouched.
--- a/packages/core/src/form/reatomFieldArray.test.ts
+++ b/packages/core/src/form/reatomFieldArray.test.ts
@@ -443,6 +443,21 @@ describe(`reset consistency (#1309)`, () => {
     ])
   })
 
+  test(`keeps dirty relevant after move`, () => {
+    const fieldArray = reatomFieldArray(['a', 'b', 'c'], {
+      name: 'resetMoveDirty.fieldArray',
+    })
+    expect(fieldArray.focus().dirty).toBe(false)
+
+    fieldArray.move(fieldArray.array()[0]!, fieldArray.array()[2]!)
+    notify()
+    expect(fieldArray.focus().dirty).toBe(true)
+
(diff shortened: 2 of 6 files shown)


Xelson:
Note on the type update in the last commit: `LinkedListMethods` is decoupled from `LinkedListAtom` so `FieldArrayAtom` can compose `LinkedListLikeAtom & Omit<LinkedListMethods, 'reset'>` with every member declared in a single place — intersecting a whole `Omit<LinkedListAtom>` duplicated the `this`-typed `extend` and broke the `isFieldArrayAtom` narrowing under `tsc -b`.

Issue #647, opened 2023-10-01
@reatom/logger: generate logs on ctx.get


Akiyamka:
Code example:
```ts
import { createCtx, atom } from '@reatom/core'
import { connectLogger } from '@reatom/logger'

const ctx = createCtx()
connectLogger(ctx)

const counterAtom = atom(0, 'counterAtom')
const doubledAtom = atom((ctx) => ctx.spy(counterAtom) * 2, 'doubledAtom')
counterAtom(ctx, 24)

console.log(ctx.get(doubledAtom)) // Should generate log
```
This code generate in logs only:
```js
42
```
But if I add subscription before reading
```ts
import { createCtx, atom } from '@reatom/core'
import { connectLogger } from '@reatom/logger'

const ctx = createCtx()
connectLogger(ctx, { skipUnnamed: false })

const counterAtom = atom(0, 'counterAtom')
const doubledAtom = atom((ctx) => ctx.spy(counterAtom) * 2, 'doubledAtom')
ctx.subscribe(doubledAtom, () => null) // Subscription added

counterAtom(ctx, 24)
console.log(ctx.get(doubledAtom))
```
Logs generated as expected
![image](https://github.com/artalar/reatom/assets/6587585/96123190-01b7-4dda-ad0d-40fa3582bbb7)

PR #1327, opened 2026-07-24
fix(core): log cold computed reads

astartsky:
I made a fix for v1001 version :)


astartsky:
https://github.com/reatom/reatom/issues/647

artalar:
I don't get it. In #647 the init log unexpectedly skips sometimes, but your fix intentionally skips logs for connected atoms.
But I can accept a change that adds an option to filter logs from the init phase of the atom.

astartsky:
@artalar Yes, you're right, I misunderstood logging concept here. Let me rethink what would be the fix for that.

PR #1331, opened 2026-07-29
fix(react): preserve atom setter type in useAtom

### LINKED ISSUE #1293: useAtom setter missing functional updater in type signature
### What happened?

`useAtom`'s setter type only allows direct values, but core `Atom.set` already supports functional updaters at runtime. Users writing `setCount(c => c + 1)` get a TS error despite correct runtime behavior.

https://github.com/reatom/reatom/blob/a7e178c5a482d0ac06542c7d5e1e11c63f642bc8/packages/react/src/hooks.ts#L22-L33

```typescript
// User code — runtime works, tsc fails
const [, setCount] = useAtom(atom(0))
setCount(c => c + 1)  // TS2345: fn not assignable to 'number'
```

Core already dispatches updater functions in `computedMiddleware`:
https://github.com/reatom/reatom/blob/a7e178c5a482d0ac06542c7d5e1e11c63f642bc8/packages/core/src/core/atom.ts#L849-L852

https://github.com/reatom/reatom/blob/a7e178c5a482d0ac06542c7d5e1e11c63f642bc8/packages/core/src/core/atom.ts#L119-L148
Type-only bug — runtime is fine.

## Fix

```diff
-    Target extends Atom<any, infer Params>
-      ? (...args: Params) => AtomState<Target>
-      : undefined,
+    Target extends Atom<any, any[]> ? Target['set'] : undefined,
```

## Approaches considered

| Approach            | Diff                                                          | Gap                                                                      |
| ------------------- | ------------------------------------------------------------- | ------------------------------------------------------------------------ |
| **Intersection**    | `((...args: Params) => State) & ((update: State) => State)`   | Loses custom `set` overloads from atom subtypes (e.g. `UrlAtom.replace`) |
| **Union spread**    | `...args: Params \| [(prev: State) => State]`                 | Same gap + fragile rest params from union tuples                         |
| **`Target['set']`** | `Target extends Atom<any, any[]> ? Target['set'] : undefined` | None                                                                     |

### Part

Package @reatom/all-settled

### Version

latest

_No response_

### Relevant log output

```shell

```

astartsky:
Closes #1293

`useAtom` now exposes the target atom's `set` type instead of reconstructing it from atom params. This enables functional updaters and preserves custom setter overloads such as `UrlAtom.set`.

Tests: `pnpm --filter @reatom/react run test`
Build: `pnpm --filter @reatom/react run build`

### DIFF
--- a/packages/react/src/hooks.ts
+++ b/packages/react/src/hooks.ts
@@ -25,9 +25,7 @@ export const useAtom: {
     options?: { subscribe?: boolean },
   ): [
     AtomState<Target>,
-    Target extends Atom<any, infer Params>
-      ? (...args: Params) => AtomState<Target>
-      : undefined,
+    Target extends Atom<any, any[]> ? Target['set'] : undefined,
     Target,
     Frame,
   ]


artalar:
Thanks!!

PR #1333, opened 2026-07-30
fix(jsx): sync aria-checked with model:checked

### LINKED ISSUE #1013: [jsx] Add aria-checked synchonization with model:checked

vianmangal:
## Summary

- synchronize `aria-checked` whenever `model:checked` updates the checked property
- cover initial, atom-driven, and input-driven changes

Fixes #1013

## Testing

- `pnpm --filter @reatom/jsx test`
- `pnpm --filter @reatom/jsx run build`

### DIFF
--- a/packages/jsx/src/index.test.tsx
+++ b/packages/jsx/src/index.test.tsx
@@ -1334,14 +1334,12 @@ test('model:field binds checkbox checked', () =>
 test('model:checked synchronizes aria-checked', () =>
   context.start(async () => {
     const checked = atom(false, 'checked')
-    const input = instance(
-      HTMLInputElement,
-      <input model:checked={checked} attr:type="checkbox" />,
-    )
+    const input = instance(HTMLInputElement, <input model:checked={checked} />)
 
     mount(parent(), input)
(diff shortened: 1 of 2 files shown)


packages/jsx/src/index.test.tsx
```diff
@@ -1331,6 +1331,34 @@ test('model:field binds checkbox checked', () =>
     expect(input.checked).toBe(true)
   }))
 
+test('model:checked synchronizes aria-checked', () =>
+  context.start(async () => {
+    const checked = atom(false, 'checked')
+    const input = instance(
+      HTMLInputElement,
+      <input model:checked={checked} attr:type="checkbox" />,
```
artalar: Wdyt to add `attr:type="checkbox"` by default too for `model:checked` and target `HTMLInputElement`?

vianmangal: Good idea. I’ll default the type to `checkbox` when `model:checked` targets an `HTMLInputElement` and no explicit type is provided, while preserving explicit types such as `radio`. I’ll add test coverage for both cases.

Issue #1334, opened 2026-07-31
withLocalStorage: a reset to `undefined` is not propagated to other tabs


rushelex:
Any value change reaches other tabs; a change to `undefined` does not.

Repro (Node, stubbed `localStorage` + a `storage` event):

```js
const KEY = 'shared-key'

const a = atom(undefined).extend(withLocalStorage(KEY))   // tab A
a.subscribe(() => {})
a.set('value')

const b = atom(undefined).extend(withLocalStorage(KEY))   // tab B, opened later
b.subscribe(() => {})
b()                       // 'value'

a.set('other')
dispatchStorageEvent(KEY) // what the browser does for the other tab
b()                       // 'other'  — fine

a.set(undefined)
dispatchStorageEvent(KEY)
b()                       // 'other'  — expected undefined
```

Cause: `toPersistRecord` writes `data: toSnapshot(state)`, and for `undefined` `JSON.stringify` drops the key, so the stored record carries no `data`. `assertPersistRecord` in the `storage` handler then rejects the record the library itself wrote, and the update is discarded — silently on 1001.3.0, as an uncaught `ReatomError` on 1001.1.0 (the handler was not wrapped in `try/catch` there).

Reproduced on 1001.1.0 and 1001.3.0.

Expected: a written record must round-trip, so a reset propagates like any other change.

PR #1335, opened 2026-07-31
test(core): failing test for cross-tab reset to undefined

rushelex:
Failing test for #1334 — no fix, just the repro.

A regular cross-tab change reaches the other tab; a reset to `undefined` does not. The test asserts both steps, so the first one documents that the mechanism works and the second one isolates the bug.

Run: `pnpm --filter @reatom/core run test:browser`

```
× localStorage propagates a reset to undefined to another tab
AssertionError: expected 'other-value' to be undefined
```


artalar:
Good stuff, will work on it in the near time.

PR #1338, opened 2026-08-03
fix(core): keepState abort — cancel the speculative body, deliver the cached state

Xelson:
## Regression (introduced by #1322)

#1322 fixed a real bug — a direct `await query()` on a persist-hydrated cache hit rejected with `AbortError: <name>.withAbort cache [#N]` — but the fix (`abortVar.set()` after `controller.abort('cache')`) broke the **other** contract of a cache hit: the *dry first run*.

The intended semantics of a hydrated hit: the computed body runs once as a pure dependency-collection pass; its deferred work — the idiomatic `await wrap(schedule(() => fetch(...)))` — checks the call's aborted controller **at flush time** and skips, so zero requests go out.

`abortVar.set()` swapped a **fresh, non-aborted** controller into the same frame slot the deferred fetch re-reads at flush time. The abort became invisible to it → the speculative fetch fired (and its result was discarded) on **every hydrated first read**. In an SSR app that's one wasted request per `withCache({ withPersist })` computed per page load, and it breaks "hydration does not refetch" assertions downstream.

## Root tension

The same `abort('cache')` must simultaneously
- **cancel the body's speculative work** — deferred effects must SEE the abort, and
- **not fail the call itself** — the returned state is the valid cached result.

One controller in one frame slot can't express both by swapping it: the two *consumers* of the abort need to be distinguished, not the controller replaced.

## Fix

Express the distinction on the shared primitive both sides already hold — the controller itself (same pattern as its existing `spawned` flag):

```ts
// ReatomAbortController
keepState?: boolean  // "cancel the in-flight work, keep the call's result"

// withCache, non-SWR hit:
controller.keepState = true
controller.abort('cache')

// withAbort, delivery side:
if (thisController.signal.aborted && thisController.keepState) {
  return state // the cached result — deliver untouched, don't abort-track it
}
```

- The aborted controller **stays in the frame** → the body's `schedule`d work still cancels before the request fires.
- The returned cached state is **not wrapped with abort tracking** → no rejection, no poisoned atom state; a direct `await` resolves with the cached value (the #1322 guarantee holds). The already-aborted controller is also not registered in `activeControllers`.
- The flag is **instance-scoped** (checked on `thisController` itself), so there's no frame-tree walk and no coupling of `withAbort` to `withCache` internals — `withCache` and `withAbort` both already depend on `methods/abortVar`, and neither imports the other.

## Tests

- **New:** `a hydrated cache hit cancels the speculative body — its scheduled fetch never fires` — red on v1001 HEAD (the fetch fires), green with this fix. Also asserts the direct-await path in the same flow.
- **Existing #1322 test** (`remount after an SSR-hydrated cache hit must render, not throw the cache AbortError`) — stays green: the two tests pin the opposite boundaries (call must not fail / body must not run).
- Full core suite: 59 files, 419 passed, 2 expected fail, no regressions.

Verified end-to-end in the app that hit the regression: its "hydration does not refetch" SSR test fails on v1001 HEAD and passes with this patch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/methods/abortVar.ts
+++ b/packages/core/src/methods/abortVar.ts
@@ -21,13 +21,6 @@ import { wrap } from './wrap'
  *   (default: `false`)
  */
 export class ReatomAbortController extends AbortController {
-  /**
-   * Marks an abort as "cancel the in-flight work, keep the call's result": the
-   * frame's state is already a settled, valid value (e.g. `withCache` serving a
-   * cache hit), so delivery-side consumers (`withAbort`) must return it
-   * untouched instead of treating the call as failed. Deferred work still
-   * observes the abort in the frame and cancels.
-   */
   keepState?: boolean
   constructor(
     public name: string,
--- a/packages/core/src/methods/abortVar.ts
+++ b/packages/core/src/methods/abortVar.ts
@@ -60,11 +60,19 @@ export class AbortVariable extends Variable<
 
   override find<Result = ReatomAbortController>(
     cb?: (payload: undefined | ReatomAbortController) => undefined | Result,
-    frame?: Frame,
+    frame: Frame = top(),
   ): undefined | Result {
     let result: undefined | Result
 
     super.find((controller) => {
+      if (
+        controller !== frame[this.name] &&
+        controller?.signal.aborted &&
+        controller.keepState
+      ) {
+        return (controller.spawned ? true : undefined) as undefined | Result
+      }
+
       result = cb ? cb(controller) : (controller as undefined | Result)
       return result !== undefined || controller?.spawned ? true : undefined
     }, frame)
--- a/packages/core/src/methods/abortVar.ts
+++ b/packages/core/src/methods/abortVar.ts
@@ -21,6 +21,14 @@ import { wrap } from './wrap'
  *   (default: `false`)
  */
 export class ReatomAbortController extends AbortController {
+  /**
+   * Marks an abort as "cancel the in-flight work, keep the call's result": the
+   * frame's state is already a settled, valid value (e.g. `withCache` serving a
+   * cache hit), so delivery-side consumers (`withAbort`) must return it
+   * untouched instead of treating the call as failed. Deferred work still
+   * observes the abort in the frame and cancels.
+   */
+  keepState?: boolean
   constructor(
     public name: string,
     public spawned: boolean = false,
--- a/packages/core/src/extensions/withAbort.ts
+++ b/packages/core/src/extensions/withAbort.ts
@@ -154,13 +154,6 @@ export let withAbort =
 
       if (hasError) throw computationError
 
-      // A `keepState` abort means "cancel the speculative body work, keep the
-      // call's result" (`withCache` serving a cache hit): deferred effects see
-      // the aborted controller in the frame and skip, while the returned state
-      // IS the valid settled result. Deliver it untouched — the abort-tracking
-      // wrap below would reject it with that same abort (poisoning the atom
-      // state / throwing on a direct await — the original #1322 bug) — and
-      // don't register the already-aborted controller as active.
       if (thisController.signal.aborted && thisController.keepState) {
         return state
       }
--- a/packages/core/src/extensions/withAbort.ts
+++ b/packages/core/src/extensions/withAbort.ts
@@ -154,6 +154,17 @@ export let withAbort =
 
       if (hasError) throw computationError
 
+      // A `keepState` abort means "cancel the speculative body work, keep the
+      // call's result" (`withCache` serving a cache hit): deferred effects see
+      // the aborted controller in the frame and skip, while the returned state
+      // IS the valid settled result. Deliver it untouched — the abort-tracking
+      // wrap below would reject it with that same abort (poisoning the atom
+      // state / throwing on a direct await — the original #1322 bug) — and
+      // don't register the already-aborted controller as active.
+      if (thisController.signal.aborted && thisController.keepState) {
+        return state
+      }
+
       activeControllers.push(thisController)
 
       if (strategy === 'first-in-win') {
--- a/packages/core/src/extensions/withCache.test.ts
+++ b/packages/core/src/extensions/withCache.test.ts
@@ -580,3 +580,65 @@ test('remount after an SSR-hydrated cache hit must render, not throw the cache A
     }),
   )
 })
+
+test('a hydrated cache hit cancels the speculative body — its scheduled fetch never fires', async () => {
+  // The counterpart boundary to the test above. A cache hit aborts the call's own
+  // controller precisely so the body's deferred work (a `schedule`d fetch, the
+  // idiomatic SSR shape) checks that aborted controller at flush time and skips —
+  // the first hydrated run is a dry dependency-collection pass, zero requests.
+  // Regressed by #1322's `abortVar.set()`: swapping a FRESH controller into the
+  // frame slot hid the abort from the deferred fetch, firing a wasted request on
+  // every hydrated first read.
+  const name = 'withCache.hitCancelsBody'
+  const storage = createMemStorage({ name, subscribe: false })
+  const withSSR = reatomPersist(storage)
+
+  let fetches = 0
+  const make = () => {
+    const cursor = atom<string | null>(null, `${name}.cursor`)
+    const query = computed(async () => {
+      const cur = cursor()
+      return await wrap(
+        schedule(async () => {
+          fetches++
+          await sleep(10)
+          return `page:${cur}`
+        }),
+      )
+    }, `${name}.query`).extend(
+      withAsyncData({ initState: 'INIT' }),
+      withCache({ withPersist: withSSR }),
+    )
+    return { cursor, query }
+  }
+
+  const snapshot = await wrap(
+    context.start(async () => {
+      const { query } = make()
+      await wrap(query())
+      await wrap(sleep(20))
+      return storage.snapshotAtom()
+    }),
+  )
+
+  await wrap(
+    context.start(async () => {
+      const { query } = make()
+      storage.snapshotAtom.set(snapshot)
+      fetches = 0
+
+      // render-style connect — the hydrated hit serves the snapshot…
+      const un = query.data.subscribe(noop)
+      await wrap(sleep(20))
+      expect(query.data()).toBe('page:null')
+      // …and the body's scheduled fetch was cancelled by the hit's abort
+      expect(fetches).toBe(0)
+
+      // the #1322 guarantee still holds: a direct call resolves with the cached value
+      expect(await wrap(query())).toBe('page:null')
+      await wrap(sleep(20))
+      expect(fetches).toBe(0)
+      un()
+    }),
+  )
+})
--- a/packages/core/src/extensions/withCache.test.ts
+++ b/packages/core/src/extensions/withCache.test.ts
@@ -11,6 +11,8 @@ import {
 import { withSearchParams } from '../routing'
 import { type Fn, noop, type Rec, sleep } from '../utils'
 import { urlAtom } from '../web'
+import { onEvent } from '../web/onEvent'
+import { withAbort } from './withAbort'
 import { withCache, type WithCacheOptions } from './withCache'
 
 test('withCache', async () => {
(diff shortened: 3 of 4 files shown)


packages/core/src/methods/abortVar.ts
```diff
@@ -21,6 +21,7 @@ import { wrap } from './wrap'
  *   (default: `false`)
  */
 export class ReatomAbortController extends AbortController {
+  keepState?: boolean
```
artalar, line 24: Definitely not. Try to replace with a variable.

Xelson:
Added `44bed0c6` — a second leak of the same `keepState` abort.

`abortVar.find` resolves the ambient controller by walking `frame.pubs`, i.e. the **dependency graph**. So the cache-hit abort does not stay inside the cached atom: any atom that *reads* it finds the aborted controller through the traverse and inherits the cancellation. `withAbort` learned about `keepState` here; `abortVar.subscribe` and `wrap` did not, which is exactly how it surfaced.

Repro (added to `withCache.test.ts`): an SSR-hydrated cursor-paginated query read by a model with its own `withAbort()` that registers an `onEvent` listener — the render reading that model rethrows `<query>.withAbort cache [#N]`. Three ingredients, removing any one hides it: reactive params on the cached atom, the listener in the reader, and a render frame rather than a bare subscribe.

The fix exempts the start frame's own controller, so the computation's deferred work still sees the abort and skips its scheduled fetch (the boundary pinned by the test above it). `spawned` marks an abort boundary, so skipping the controller stops the traverse there instead of walking past it.

Not the root cause: an abort scope is a property of the dynamic extent, but it is resolved through the dependency graph, and the two only coincide while a dependency is being computed inside its dependent — caching is what makes them diverge. A live controller belonging to a dependency can still become a reader's scope; only the `keepState`-aborted case is closed here. The principled fix is an explicit scope link on the frame, which would also retire `spawned` and `_findReactiveStartIndex` — filed as a separate concern, since it changes what a shared computed's scope *is*.

Perf: `Variable.find` is not on the reactive hot path (0 calls across 100k updates over a 20-deep graph). Interleaved A/B on `bench_computed` shows no effect — the sign flips between pairs and the within-condition spread exceeds the difference.

Xelson:
## RFC: resolve the abort scope from the execution stack, not the dependency graph

Follow-up to `44bed0c6` — that commit patches the symptom; this proposes removing the mechanism that produces the whole bug class. Prototype included (3 files, +54/−22), every claim below is measured, not argued.

### Problem

An abort scope is a property of the **dynamic extent** — "who is executing me". `abortVar.find` resolves it by walking `frame.pubs` — the **dependency graph** — "whose data am I reading". The two coincide only while a dependency is computed inside its dependent. Caching/hydration is exactly what makes them diverge: a dependent atom then finds a *finished* computation's controller on a frame that outlived it, and inherits a cancellation it was never inside of.

Three special cases each patch one consequence of this conflation:

- `spawned` — manually stops the walk at a render boundary
- `_findReactiveStartIndex = 1` — reactive atoms must not inherit their actualization cause (`pubs[0]`), so the walk skips it… and reaches the same controllers anyway, through dependencies
- `keepState` in the *lookup* — "this abort doesn't count". `withAbort` learned the flag; `abortVar.subscribe` and `wrap` didn't, which is precisely the bug `44bed0c6` fixes

The failure mode of flag-based exemptions is structural: every consumer of the ambient abort must learn every flag, and each one that doesn't is the same bug again.

### What was tried first (all measured, all rejected)

| variant | failures | what it proved |
|---|---|---|
| remove the dep-walk entirely | 5 | the walk has one real job: reaching the **enclosing action's** controller via `dep → pubs[0]` |
| explicit `scope` field on the frame | 8 | frames are **reused** across calls — one link per frame cannot represent a per-call extent |
| follow `pubs[0]` (cause) transitively | same 5 | the cause chain adds nothing over removing the walk |
| structural rule (`pubs[0]` ok, `pubs[1..]` no) + keepState | 1 | close — but `pubs[1..]` mixes data deps with the atom's own infra frames (`._subscribe`) |

### The model (prototype attached)

Four principles, each pinned by a failure that appears when it's removed:

1. **Call scope = the execution stack.** `abortVar.find` walks `STACK` — which the runtime already maintains (`top()` reads it) — instead of the dependency graph. No flags in the lookup at all.
2. **Capture at creation.** `wrap` resolves the ambient controller **once**, when the stack is live, and closes over it; deferred checks use the captured reference. This is what `wrap`'s own JSDoc already promises by citing TC39 `AsyncContext.Snapshot.wrap` — the implementation just didn't do it. (Deferred re-resolution through an explicit frame was the *second* leak path in the PR's bug — the probe showed zero leaks through `find` while `throwIfAborted(frame)` still delivered the cancelled controller.)
3. **Continuations restore the full snapshot, with replace semantics.** `seal` used to restore a single frame — a nested `wrap` created inside a continuation then captured a near-empty stack (`@reatom/react`'s `aborts on unmount` catches exactly this: a polling loop that must die with its component survived). And restore must **replace**, not append: appending duplicated the scheduler's ambient frame, +1 stack frame per `await`, a linear leak (OOM'd a 30k-iteration benchmark).
4. **Lifetime owns the call.** An unsubscribe aborts the *current run's* controller (reason preserved), so stack-resolved consumers hear it with no graph reachability. The `effect` bridge closes the existing `// TODO … need to link abortVar to unsubscribe`.

### What `keepState` becomes

Not removed — **shrunk to its correct size**. Removing it under the new model still fails the two delivery tests (`#1322` remount + the scheduled-fetch-never-fires one): it answers a *delivery* question — "should this abort fail the call?" — which no scope-resolution change can answer. But it disappears from the lookup entirely. Before: one writer, unbounded readers, every unlearned reader a new bug. After: one writer (`withCache`), one reader (`withAbort`), both about the same call.

### Results

| suite | result |
|---|---|
| core unit | 424 ✓ |
| core browser | 115 ✓ |
| react (chromium) | 36 ✓ |
| jsx | 103 ✓ |
| zod | 35 ✓ |

Also retired: the dep-graph walk in `abortVar` (no call site reaches it any more), `_findReactiveStartIndex` for aborts, and the `44bed0c6` lookup-skip (subsumed).

### Performance (measured, honest)

The sync hot path is untouched — `bench_computed` is unaffected (the abort lookup is not on it: 0 `find` calls across 100k updates of a 20-deep graph). The async micro-path pays for two array allocations per `await` (creation snapshot + restore save):

| load | upstream | prototype | per-op |
|---|---|---|---|
| `await wrap()` ×30k, shallow | 248ms | 280–348ms | +1–3µs/await |
| `await wrap()` ×5k, depth 10 | 112ms | 153–161ms | +8µs/await |
| `wrap(fn)` create+call ×200k | 49ms | 76–100ms | +0.1–0.25µs |

That's +30–50% on the µs-scale overhead — invisible behind real I/O, but real. The canonical remedy is the standard AsyncContext implementation technique: a **persistent (immutable cons-list) stack**, making snapshot O(1) pointer capture and restore O(1) pointer swap, zero allocations. Cost: `STACK` is directly indexed (`STACK[STACK.length - 2]`, …) in ~15 places in `atom.ts`, so that's the real refactor this RFC would entail.

### Open questions

1. Should a computed inherit the call scope of whoever actualizes it? Today the answer is accidental (yes, but only via dependency reachability, only when a dep happened to actualize under the same action). The stack model makes it a real decision.
2. Migration: a dual-resolve period — resolve both ways, `console.warn` in dev on divergence — surfaces ecosystem reliance on graph reachability before the walk is deleted.
3. The persistent-stack refactor (perf) — land with the model or after it.

<details>
<summary>Prototype diff (3 files, +54/−22)</summary>

```diff
diff --git a/packages/core/src/methods/abortVar.ts b/packages/core/src/methods/abortVar.ts
index e6db8877..4248189b 100644
--- a/packages/core/src/methods/abortVar.ts
+++ b/packages/core/src/methods/abortVar.ts
@@ -1,5 +1,6 @@
 import {
   bind,
+  STACK,
   type Frame,
   type GAction,
   top,
@@ -60,24 +61,25 @@ export class AbortVariable extends Variable<
 
   override find<Result = ReatomAbortController>(
     cb?: (payload: undefined | ReatomAbortController) => undefined | Result,
-    frame: Frame = top(),
+    frame?: Frame,
   ): undefined | Result {
-    let result: undefined | Result
+    if (frame !== undefined) return super.find(cb as never, frame)
+
+    for (let i = STACK.length - 1; i >= 0; i--) {
+      let current = STACK[i]!
+      let controller = current['var#abort']
+      let result = cb ? cb(controller) : (controller as undefined | Result)
 
-    super.find((controller) => {
       if (
-        controller !== frame[this.name] &&
-        controller?.signal.aborted &&
-        controller.keepState
+        result !== undefined ||
+        controller?.spawned ||
+        current.atom === this.spawn
       ) {
-        return (controller.spawned ? true : undefined) as undefined | Result
+        return result
       }
+    }
 
-      result = cb ? cb(controller) : (controller as undefined | Result)
-      return result !== undefined || controller?.spawned ? true : undefined
-    }, frame)
-
-    return result
+    return undefined
   }
 
   declare createAndRun: GAction<
diff --git a/packages/core/src/methods/effect.ts b/packages/core/src/methods/effect.ts
index 46a07d29..f2cffd49 100644
--- a/packages/core/src/methods/effect.ts
+++ b/packages/core/src/methods/effect.ts
@@ -76,10 +76,22 @@ export let effect = <T>(cb: () => T, name?: string) => {
     //   }
     //   return
     // }
-    if (_read(target.subscribe as Action)?.['var#abort']?.signal.aborted) {
+    let life = _read(target.subscribe as Action)?.['var#abort']
+    if (life?.signal.aborted) {
       return
     }
 
+    // lifetime owns the call: unsubscribe cancels the current run, so consumers
+    // resolving their scope from the stack hear it without any graph reachability
+    let run = top()['var#abort']
+    if (life && run) {
+      life.signal.addEventListener(
+        'abort',
+        () => run.abort(life.signal.reason),
+        { signal: run.signal },
+      )
+    }
+
     let res = cb()
     if (res instanceof Promise) {
       res.catch((error) => {
diff --git a/packages/core/src/methods/wrap.ts b/packages/core/src/methods/wrap.ts
index d86856dd..52615860 100644
--- a/packages/core/src/methods/wrap.ts
+++ b/packages/core/src/methods/wrap.ts
@@ -56,17 +56,27 @@ export let wrap: {
   frame = top(),
 ): T extends Fn ? ReturnType<T> : Promise<Awaited<T>> => {
   let { root } = frame
+  let snapshot = STACK.slice()
 
   if (typeof target === 'function') {
-    abortVar.throwIfAborted(frame)
+    let controller = abortVar.get()
+    if (controller?.signal.aborted) throw controller.signal.reason
 
     return function wrap(...params: any) {
-      return frame.run(() => {
-        if (root !== frame.root) throwAbort('context reset')
-        abortVar.throwIfAborted(frame)
-        // @ts-expect-error
-        return target(...params)
-      })
+      let saved = STACK.slice()
+      STACK.length = 0
+      STACK.push(...snapshot.slice(0, -1))
+      try {
+        return frame.run(() => {
+          if (root !== frame.root) throwAbort('context reset')
+          if (controller?.signal.aborted) throw controller.signal.reason
+          // @ts-expect-error
+          return target(...params)
+        })
+      } finally {
+        STACK.length = 0
+        STACK.push(...saved)
+      }
     } as any
   }
 
@@ -82,9 +92,17 @@ export let wrap: {
       abortSubscription.unsubscribe()
     }
 
-    queueMicrotask(() => void STACK.push(frame))
+    let saved: Array<Frame>
+    queueMicrotask(() => {
+      saved = STACK.slice()
+      STACK.length = 0
+      STACK.push(...snapshot)
+    })
     cb()
-    queueMicrotask(() => void STACK.pop())
+    queueMicrotask(() => {
+      STACK.length = 0
+      STACK.push(...saved!)
+    })
   }
 
   let aborted = false
```

</details>

<details>
<summary>Async micro-benchmark used for the A/B (not in the repo)</summary>

```ts
import { action, context, wrap } from '../src/index'

context.start(async () => {
  const N = 30_000
  const shallow = action(async () => {
    for (let i = 0; i < N; i++) await wrap(Promise.resolve(i))
  }, 'bench.shallow')

  const leaf = action(async () => {
    for (let i = 0; i < 5_000; i++) await wrap(Promise.resolve(i))
  }, 'bench.leaf')
  let chain = leaf
  for (let d = 0; d < 8; d++) {
    const next = chain
    chain = action(async () => await wrap(next()), `bench.d${d}`)
  }

  const F = 200_000
  const fnbench = action(() => {
    let acc = 0
    for (let i = 0; i < F; i++) {
      const f = wrap((x: number) => x + 1)
      acc = f(acc)
    }
    return acc
  }, 'bench.fn')

  const time = async (label: string, fn: () => any) => {
    const runs: number[] = []
    for (let r = 0; r < 5; r++) {
      const t = performance.now()
      await wrap(fn())
      runs.push(performance.now() - t)
    }
    runs.sort((a, b) => a - b)
    console.log(`${label}: median=${runs[2]!.toFixed(1)}ms min=${runs[0]!.toFixed(1)}ms`)
  }

  await time('await-wrap shallow x30k   ', () => shallow())
  await time('await-wrap deep(9)  x5k   ', () => chain())
  await time('wrap(fn) create+call x200k', () => fnbench())
  process.exit(0)
})
```

</details>

PR #1341, opened 2026-08-12
fix(core): log computed initialization

### LINKED ISSUE #647: @reatom/logger: generate logs on ctx.get

astartsky:
## Summary

- Log the first computed evaluation consistently, whether triggered by a cold read or by a subscription.
- Keep repeated reads of unchanged state from producing duplicate log entries.
- Add regression tests for both initialization paths.

## Context

[#647](https://github.com/reatom/reatom/issues/647) reports that reading an unconnected computed does not produce the expected logger entry.

The previous approach in [#1327](https://github.com/reatom/reatom/pull/1327) treated connected and disconnected computeds differently. This revision removes that distinction: computed initialization is now logged regardless of how the computation becomes active, while the existing unchanged-state check prevents duplicate logs.

## Testing

- `pnpm --filter @reatom/core test`

Fixes #647

### DIFF
--- a/packages/core/src/methods/connectLogger.test.ts
+++ b/packages/core/src/methods/connectLogger.test.ts
@@ -1,6 +1,6 @@
 import { expect, test, vi } from 'test'
 
-import { sleep, wrap } from '..'
+import { peek, sleep, wrap } from '..'
 import { atom, computed, EXTENSIONS, notify, withActions } from '../core'
 import { connectLogger, log } from './connectLogger'
 import { getStackTrace } from './getStackTrace'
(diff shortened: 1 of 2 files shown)


artalar:
We shouldn't log cold atoms reads, for all atoms overview the user should use the states overview tab in the admin devtools

PR #1353, opened 2026-08-27
test(core): reproduce phantom Error("undefined") on nullish rejection reason

AgarkovRoman:
## Summary

`withAsync`'s `onReject` and `withAsyncStatus`'s `getStatusError` disagree on how to treat a nullish (`undefined`/`null`) rejection reason. `onReject` only skips setting `error` when `isAbort(err)` is true; `getStatusError` additionally skips it when the rejection is `== null`. So a nullish rejection reason turns into a fabricated `Error("undefined")` via `error()`, while `status().error` correctly stays empty for the same rejection.

## Why this isn't just theoretical

`AbortController.abort()` with no argument is supposed to default `signal.reason` to a spec `AbortError`. Not every polyfill implements that default correctly — we've observed `signal.reason` staying `undefined` after a no-argument abort on React Native/Hermes. Anything awaiting a signal-driven promise (e.g. `fetch(url, { signal })`) on such a runtime can reject with `undefined` on an ordinary abort, and `withAsync` turns that into a phantom, misleading error.

## What this PR adds

Three failing regression tests (`withAsync.test.ts`, `withAsyncData.test.ts`, `withAsyncStatus.test.ts`) that mock a target to reject with `undefined` and assert `error()`/`status().error`/`data()` stay empty, consistent with how an abort is handled elsewhere. No fix is included — reporting the bug with tests first, open to discussing the right fix.

## Proposed fix (not applied here)

Align `onReject`'s guard with `getStatusError`'s, in `packages/core/src/async/withAsync.ts`:

```diff
-  if (!isAbort(err)) {
+  if (!isAbort(err) && err != null) {
     error.set((err = parseError(err)))
   }
```

### DIFF
--- a/packages/core/src/async/withAsync.test.ts
+++ b/packages/core/src/async/withAsync.test.ts
@@ -176,6 +176,23 @@ test('abort rejection settles without calling onReject', async () => {
   expect(onSettle).toHaveBeenCalledTimes(2)
 })
 
+test('rejection with a nullish reason does not produce a phantom Error("undefined")', async () => {
+  const fetchSmth = action(async () => 'real', 'phantomAbortError').extend(
+    withAsync(),
+  )
+  const unmock = mock(fetchSmth, () => Promise.reject(undefined))
+
+  try {
+    try {
+      await wrap(fetchSmth())
+    } catch {}
+
+    expect(fetchSmth.error()).toBeUndefined()
+  } finally {
+    unmock()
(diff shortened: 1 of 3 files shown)


artalar:
Reatom can't handle nullish errors, it is expected that an error can't be nullish, so it transforms to `Error("undefined")` in the catch case.

